PR #259, opened 2023-01-27
JCL-67: e2e on PodSpaces

timea-solid:
Preliminary draft work towards separation of mock services.


acoburn (approved):
This looks great -- there are a few minor things related to try-with-resources and streams, but we can fix those up separately. Also, I'm not a big fan of global state structures like the `State` class, but this is certainly fine for now.

PR #266, opened 2023-02-02
RDF wrapping

langsamu:
Fourth attempt at introducing RDF wrapping.

This is an almost exact rework of #250 (which details the changes) but using the wrapping library extracted in inrupt/commons-rdf-wrapping#8.

~~This will not work until the other library goes to Maven Central.~~
**This is worth reviewing now that `inrupt-rdf-wrapping` has been released.**


acoburn (approved):
This is great. I would like to have @timea-solid merge #259 first, since that's the larger PR and it looks like there may be some conflicts

PR #277, opened 2023-02-08
JCL-67: CI for live testing

timea-solid:
This is based on 2 environments set up on GitHub secrets: ESS PodSpaces and ESS DevNext. Each of them defines:
* INRUPT_TEST_WEBID
* INRUPT_TEST_CLIENT_ID
* INRUPT_TEST_CLIENT_SECRET
* INRUPT_TEST_PUBLIC_RESOURCE_PATH

These are then read in the tests through the CI.
The CI is taken from a [Javascript project](https://github.com/inrupt/solid-client-js/blob/main/.github/workflows/e2e-browser.yml) (builds on top of their learnings). For example, DevNext tests are not gating on purpose.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,7 @@ version: 2
 updates:
   - package-ecosystem: "maven"
     directory: "/"
+    target-branch: "JCL-67-ci"
     schedule:
       interval: "weekly"
   - package-ecosystem: "github-actions"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,7 +2,6 @@ version: 2
 updates:
   - package-ecosystem: "maven"
     directory: "/"
-    target-branch: "JCL-67-ci"
     schedule:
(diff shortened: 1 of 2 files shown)


acoburn (approved):

NSeydoux (changes requested):
The `experimental` flag on the matrix should be used in order to determine whether the job failing should fail the workflow or not: https://github.com/inrupt/solid-client-access-grants-js/blob/57f050215d6f5c3d7459c4b49e2d59bacfaec24b/.github/workflows/e2e-node.yml#L20. In other words, that's an ad-hoc configuration we use to set a standard value, rather than being a standard value in the first place.

NSeydoux (approved):
Looks good, although maybe a comment on the Maven flags used to distinguish between unit tests and integration ones would be helpful? And same, can you make sure that there is either a comment or a commit message that explains why dependabot PRs skip integration tests, otherwise that'll definitely confuse future us.

PR #278, opened 2023-02-08
JCL-207: Add support for access grants

acoburn:
This follows the existing pattern for Session usage:

```java
var session = AccessGrant.ofAccessGrant(oidcSession, accessGrant1, accessGrant2);
```

Internally, the appropriate access grant will be used for the relevant request. The wrapped `oidcSession` object manages OIDC-related interactions.

This modifies the signature of `Session::getCredential` to include a `URI` parameter.


timea-solid (approved):
I cannot say I understand access grants from the code only. I would like to read some documentation about it.

PR #293, opened 2023-02-14
JCL-253: cleanup

timea-solid:
This resolves open feedback on PR: https://github.com/inrupt/solid-client-java/pull/259#pullrequestreview-1285176279
And also adds integration tests in the big table over at https://inrupt.github.io/solid-client-java/


acoburn (commented):
You should be able to take my suggestions and apply them to many of the other changes here.

acoburn (approved):

PR #299, opened 2023-02-17
JCL-227: Header decorator

acoburn:
This adds a mechanism for adding headers to requests in the high-level client API. The existing API stays the same with an additional option added. For example:

```java
MyCustomType obj = client.read(uri, customHeaders, MyCustomType.class);
```

In addition, a header collection can be added to all requests when constructing the client:

```java
SolidClient client = SolidClient.getClientBuilder().headers(customHeaders).build();
```

The `Builder` mechanism is new and replaces the (inconsistent) `SolidClient.of(...)` methods.

In addition, a default `User-Agent` header is now added to each request (this can be overridden by a developer). By default, this is `InruptJavaClient/{version}`.

### DIFF
--- a/api/src/main/java/com/inrupt/client/Request.java
+++ b/api/src/main/java/com/inrupt/client/Request.java
@@ -29,6 +29,7 @@ import java.net.URI;
 import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.ArrayList;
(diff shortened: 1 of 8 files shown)


edwardsph (approved):
This looks exactly right to me - nice work.
Are tests part of another PR? - I had to ask!

PR #318, opened 2023-02-27
JCL-73: Simplify project setup as a step to simplifying the docs

acoburn:
This is a stepping stone to improving our docs by first making the initial project setup easier for developers.

This introduces an archetype submodule so that, with a single command, a developer can create a new project.


timea-solid (approved):
Having more ways to work and learn the code is never bad at the end of the day.

PR #319, opened 2023-02-28
JCL-185: Improved readme & new code of conduct

timea-solid:
This is an improved README based of the Javascript repos readme. 
I also took the liberty to add the same code of conduct reference like in the javascript repos.


acoburn (approved):
This is great -- just a few minor typographical suggestions

PR #321, opened 2023-03-01
Small code improvement

timea-solid:
Using URIBuilder instead of complicated String concats in AuthenticationScenario e2e test.


acoburn (approved):
Nice!

PR #322, opened 2023-03-02
JCL-266: fix of jena code example

timea-solid:
Ben asked me before about the Jena example in the doc. It turns out that that part was copied from rdf4j and never updated according to Jena. I think now it should do.


langsamu (changes requested):
`Model#add` has no overload that takes three strings.

langsamu (approved):

PR #323, opened 2023-03-02
JCL-237: Issue templates

timea-solid:
Quarkus inspired and mixed a bit with the javascript ones too. Most changes are in the request feature template.
@NSeydoux what do you say? would you also consider them?


langsamu (commented):
Some suggestions, nothing critical.

acoburn (approved):

PR #335, opened 2023-03-06
JCL-268: Add client identifier documents to site

acoburn:
This adds the client identifier documents to the site so that they don't get inadvertently removed in the future.


timea-solid (approved):
We are also hosting the Android one?

PR #337, opened 2023-03-07
JCL-270: Include archetype in release process

acoburn:
This includes the archetype in the deployment/release process.


davidboweninrupt (approved):
Rubber stamp

PR #339, opened 2023-03-08
JCL-271: remove 'what you will need'

timea-solid:
A feedback point was “every single page contains what you need  -> on getting started is enough, cause you build on top of that anyway.”
This is removing that section in following pages.


acoburn (commented):
Let's hold off on this change. We have to assume that developers will not proceed linearly through these pages or that they start at the beginning. Repetition here can be useful.

PR #342, opened 2023-03-08
JCL-272: Add a legacy rdf4j module

acoburn:
While the Android API 33 supports RDF4J v4.x, earlier API levels do not. For example, an Android phone v11 (API 30) cannot work with our RDF4J module.

This PR adds a new module, called `inrupt-client-rdf-legacy`. It does not include body handlers/publishers; only the RDF parsing capability, working with Java 8


timea-solid (commented):
I wonder: can we not simply use the [RdfServices tests](https://github.com/inrupt/solid-client-java/blob/main/test/src/main/java/com/inrupt/client/test/RdfServices.java) from the test module here too.

timea-solid (approved):
Ah nevermind my previous comment, I was thinking about the shared tests but these tests are not share.

PR #343, opened 2023-03-08
JCL-271: improved data modeling section

timea-solid:
Worked on this feedback "next I tried data modeling - looked promising at first”
    “I copied the classes in but without imports, I did not know if it is IRI from Jena or...”
    “At example 4  - it was building and compiled but when I ran it I had an” error -> see screenshot.”

P.s. I think it did not run because maybe of the OpenIdConnect instead of OpenIdConfig.


acoburn (commented):
Looks good, just a few minor comments

acoburn (approved):

PR #350, opened 2023-03-10
JCL 273: Changing 'testing' Client Id for Springboot App

bncummings:
I noticed that this document was originally quite an abstract example of a clientId document. I'm not sure if the changes needed justify having an example client Id document exclusively for our springboot app.


acoburn (commented):
I spent a little time seeing if we could just re-use the existing test client, but to no effect. Let's use this one, but I'd suggest a little minor clean-up first. I'd also suggest that we have the word `test` in the name of this file: e.g.: `springboot-test.jsonld`

acoburn (approved):

PR #358, opened 2023-03-13
JCL-272: Use older commons-codec methods

acoburn:
Android appears to use a much older version of commons-codec, and it is not possible to just upgrade to a newer version. With this PR, we continue to use commons-codec, but only the methods that have been around forever.


timea-solid (approved):
We are not bringing any security consideration with going so far back on the version, are we?

PR #360, opened 2023-03-16
JCL-297: springBoot version 2

timea-solid:
With auth


acoburn (approved):
This is great. I tested the auth and it works as advertised.

timea-solid:
Merged because we only had a CVE-2023-1370 which I will look into separately.

PR #368, opened 2023-03-23
JCL-275: Better support for NonRDF resources

acoburn:
This is an alternative to #367 (an improvement IMO)

There are a number of key features of this PR.

In the `com.inrupt.client` package:

* The type hierarchy is more rational: the lowest level resource type is the `Resource` (this becomes an `interface` and is independent of the RDF wrapping, because java does not support multiple inheritance of concrete classes)
* The former `Resource` class is now called `RDFSource` and implements `Resource`. This class extends the RDF Wrapping code.
* There is a new class called `NonRDFSource` which is also implements `Resource`

Classes that previously extended `Resource` will now need to extend `RDFSource`

In the `com.inrupt.client.solid` package:
* There is a `SolidBinary` class that extends `NonRDFSource` in the same way that `SolidResource` extends `RDFSource`. (We could consider some more significant renaming here, but I'm trying to have as light a touch as possible).

Otherwise, the high-level API looks the same.

There is no change to the low-level API.

*Note*: I have not added any Javadocs because I would like an initial review of this API before spending time on docs.

### DIFF
--- a/api/src/main/java/com/inrupt/client/NonRDFSource.java
+++ b/api/src/main/java/com/inrupt/client/NonRDFSource.java
@@ -20,24 +20,45 @@
  */
 package com.inrupt.client;
 
-import static org.junit.jupiter.api.Assertions.*;
-
-import com.inrupt.client.spi.RDFFactory;
-
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
 import java.net.URI;
+import java.util.Objects;
 
-import org.apache.commons.rdf.api.RDF;
-import org.junit.jupiter.api.Test;
+public class NonRDFSource implements Resource {
 
-class ResourceTest {
+    private final URI identifier;
+    private final String contentType;
+    private final InputStream entity;
 
-    static final RDF rdf = RDFFactory.getInstance();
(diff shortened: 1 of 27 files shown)


timea-solid (approved):
I like the new names and am very much ok with SolidBinary too.
Besides javadoc we will probably need to check the docs in package-infos

langsamu (commented):
I like the new hierarchy:

```
Resource
    RDFSource
        SolidResource
    NonRDFSource
        SolidBinary
```

As noted inline, I'm missing an orthogonal marker interface for SolidBinary and SolidResource.

I have a problem with content-type handling on `RDFSource`. I think the choices I criticized inline stem from the difference in construction of RDF/NonRDF resource classes.

Compare
```java
RDFSource(final URI identifier, final Dataset dataset)
```
with
```java
NonRDFSource(final URI identifier, final String contentType, final InputStream entity)
```

I think RDFSources should also take `String contentType` and `InputStream entity`, which would facilitate the kind of contentType handling I mentioned inline, and would make the two branches nice and similar.

Is it reasonable for the client to always pass in an input stream (and the rest), which would put the responsibility of constructing a dataset into the hands of `RDFSource`?

PR #370, opened 2023-03-24
JCL-278: Quarkus Webapp Style 

bncummings:
I'm not a graphic designer by any stretch, but hope this matches the 'Inrupt theme' a bit more.
Screenshot 2023-03-24 at 09 45 16


timea-solid (changes requested):
Looks really good!
The only thing we should change is the name Java SDK -> maybe Java Client Libraries Demo

timea-solid (approved):

PR #376, opened 2023-03-24
JCL-308: Add convenience function for resource deletion

acoburn:
This adds a convenience function for deleting resources, using just a URI.


timea-solid (approved):
This makes sense with the new Classes.

PR #377, opened 2023-03-24
JCL-307: Returned modified object

acoburn:
~I would like to add some more test cases here before merging, but this provides the basic outline of the implementation.~

Done


timea-solid (commented):
It looks ok in general.

langsamu (approved):

PR #379, opened 2023-03-25
JCL-303: Minor updates to the security policy

acoburn:
This removes references to the JS libraries in this security policy and fixes some minor typographical issues.


langsamu (approved):

timea-solid (approved):
Nic AS mentioned that there was an initiative to reference libs between eachother. Maybe I was a bit too fast here. I did take the Microsoft example to on point :)
Anyway, I am fine with either way.

PR #392, opened 2023-03-29
JCL-297: Add explicit version for spring boot plugin

acoburn:
This ensures that we manage the version of the springboot plug-in. (otherwise, maven just uses the latest version)


langsamu (approved):

timea-solid (approved):

timea-solid:
good point & catch

PR #400, opened 2023-04-04
JCL-317: Deprecate use of stream in return types

acoburn:
This deprecates the `SolidContainer::getContainedResources` method and adds a new `SolidContainer::getResources` method.

The deprecated method returns a `Stream`, while the new one returns an immutable `Set`. This will make it easier to work with in Kotlin-based systems.

### DIFF
--- a/archetypes/java/src/main/resources/archetype-resources/src/main/java/SolidApp.java
+++ b/archetypes/java/src/main/resources/archetype-resources/src/main/java/SolidApp.java
@@ -29,9 +29,8 @@ public class SolidApp {
         session.getPrincipal().ifPresent(webid -> {
             try (final var profile = client.read(webid, WebIdProfile.class)) {
                 profile.getStorage().stream().findFirst().ifPresent(storage -> {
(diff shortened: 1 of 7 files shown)


timea-solid (approved):
Should we change the examples & archetype already now?
* [archetype](https://github.com/inrupt/solid-client-java/blob/520fb69c352c3c783f7bbddf12867f462b313797/archetypes/java/src/main/resources/archetype-resources/src/main/java/SolidApp.java#L33)
* [quarkus example](https://github.com/inrupt/solid-client-java/blob/1a0236dca321dd1618e62d499c3a48c0c88b3b93/examples/webapp/src/main/java/com/inrupt/client/examples/webapp/SolidStorage.java#L76)
* [cli example](https://github.com/inrupt/solid-client-java/blob/520fb69c352c3c783f7bbddf12867f462b313797/examples/cli/src/main/java/com/inrupt/client/examples/cli/SolidApp.java#L90)

timea-solid (approved):

PR #401, opened 2023-04-04
JCL-321: Use com.inrupt.client for Maven groupId

acoburn:
This changes the groupId of the client libraries from `com.inrupt` to `com.inrupt.client`

### DIFF
--- a/access-grant/pom.xml
+++ b/access-grant/pom.xml
@@ -2,7 +2,7 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>com.inrupt</groupId>
+    <groupId>com.inrupt.client</groupId>
     <artifactId>inrupt-client</artifactId>
     <version>1.0.0-SNAPSHOT</version>
   </parent>
@@ -20,7 +20,7 @@
 
   <dependencies>
     <dependency>
-      <groupId>com.inrupt</groupId>
+      <groupId>com.inrupt.client</groupId>
       <artifactId>inrupt-client-api</artifactId>
       <version>${project.version}</version>
     </dependency>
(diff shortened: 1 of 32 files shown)


hzbarcea:
Looks good. Another (arguably better) option would be to leave `groupId=com.inrupt` in the root project and create a `./parent` subproject (with `groupId=com.inrupt.client`) and move all dependency management there. Then all modules would have this as the <parent> project with a `<relativePath>../parent</relativePath>`.

A bit cleaner, a bit more work. This works too, for all practical purposes almost the same thing.

hzbarcea (approved):

hzbarcea:
[Documentation](https://docs.inrupt.com/developer-tools/java/client-libraries/getting-started/step1/) will need some updates.

acoburn:
I had also been considering the creation of an independent "parent" project, living in its own repo. Then that parent can be consumed by all Java projects in Inrupt, including these client library. That change would basically involve moving most of the current root pom.xml into a separate repo and then just consuming that from here.

acoburn:
@hzbarcea the pattern you describe is something I see quite a bit elsewhere. What do you see as the advantage of having a special `./parent` module over having everything in a single root `pom.xml`? The main advantage I see is that it means the project can put all the dependency management and plugin configuration in a single place, without mixing that with other project metadata (scm, modules, developers, etc). And do you see that as preferable over having, for example, a multi-project parent POM, a-la Apache's parent POM that is used by most projects in the ASF?

hzbarcea:
> @hzbarcea the pattern you describe is something I see quite a bit elsewhere. What do you see as the advantage of having a special `./parent` module over having everything in a single root `pom.xml`? The main advantage I see is that it means the project can put all the dependency management and plugin configuration in a single place, without mixing that with other project metadata (scm, modules, developers, etc). And do you see that as preferable over having, for example, a multi-project parent POM, a-la Apache's parent POM that is used by most projects in the ASF?

You are correct. My personal preference is to have that separation, but I recognize that it's quite subjective.

hzbarcea:
Pressed <enter> too fast. The second topic you mentioned (separate and orthogonal): having a parent project for all java projects, is imho more interesting and maybe more important. That would ensure a consistent dependency management across all (or most) java projects. It would make many things easier (faster, cheaper, better) in terms of versions baseline management, upgrades, testing, deviations from baseline. It could be done at any time though, but the earlier the better.

timea-solid (approved):

PR #402, opened 2023-04-05
JCL-322: added a Objects.requireNonNull

timea-solid:
Also added a test.

### DIFF
--- a/core/src/main/java/com/inrupt/client/core/DefaultClient.java
+++ b/core/src/main/java/com/inrupt/client/core/DefaultClient.java
@@ -58,6 +58,7 @@ public final class DefaultClient implements Client {
     }
 
     DefaultClient(final HttpService httpClient, final Session session) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/inrupt/client/core/DefaultClient.java
```diff
@@ -63,6 +64,7 @@ public final class DefaultClient implements Client {
 
     @Override
     public Client session(final Session session) {
+        Objects.requireNonNull(session, "Session may not be null!");
```
acoburn, line 67: I would be inclined to omit the check here and, instead, put it in the ctor:

https://github.com/inrupt/solid-client-java/blob/6acfe3a972bea515496d442d1f2395a758f6fe50/core/src/main/java/com/inrupt/client/core/DefaultClient.java#L60-L61

acoburn (approved):
This will work, but I'd suggest a slight modification

PR #403, opened 2023-04-06
JCL-299: verify method

timea-solid:
@jholleran Could you please take a look at the verify(...) method I added. I understand there are options in the call. What can they be? is it different from access grant to access request? 
Any help is welcome. 
Conclusion: we took out options from call.


access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
```diff
@@ -181,6 +182,41 @@ public CompletionStage<AccessGrant> issue(final URI type, final URI agent, final
...
     }
 
+    /**
+     * Verify an access grant or request.
+     *
+     * @param accessGrant the access grant to verify
+     * @param options the options to which this credential applies
+     * @return the next stage of completion containing the resulting credential
+     */
+    public CompletionStage<VerificationResponse> verify(final AccessGrant accessGrant, final Set<String> options) {
```
jholleran, line 192: The options here should really be a Map<String, String>. A client will need to be able pass in key value pairs. See the example in: https://w3c-ccg.github.io/vc-api/verifier.html#tag/Presentations/operation/verifyPresentation.
For example:
```
"domain": "example.com",
"challenge": "d436f0c8-fbd9-4e48-bbb2-55fc5d0920a8"
```

Having said that the Inrupt Access Grant Service currently does not use the `options` passed in by a client. Its only there to align with the VC-API spec. A external client passing in a `domain` or `challenge` does not make sense at this API level. It is really there for lower level internal API calls between an VC Coordinator (Inrupt Access Grant Service) and a generic VC Verifier. 

So leaving the `options` parameter (as a Map) here is ok as an expansion point but it could also be cleaner to remove it as I can't see it been used at this API level.

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
```diff
@@ -320,6 +356,26 @@ AccessGrant processVerifiableCredential(final InputStream input, final Set<Strin
         }
     }
 
+    String prepareVerifyPayload(final InputStream input, final Set<String> validTypes, final Set<String> options) throws IOException {
```
jholleran, line 359: Who calls this?

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
```diff
@@ -181,6 +182,41 @@ public CompletionStage<AccessGrant> issue(final URI type, final URI agent, final
...
+     *
+     * @param accessGrant the access grant to verify
+     * @param options the options to which this credential applies
+     * @return the next stage of completion containing the resulting credential
+     */
+    public CompletionStage<VerificationResponse> verify(final AccessGrant accessGrant, final Set<String> options) {
+        return v1Metadata().thenCompose(metadata -> {
+            
+            final Map<String, Object> presentation = new HashMap<>();
+            presentation.put(VERIFIABLE_CREDENTIAL, accessGrant);
```
jholleran, line 196: The `/verify` endpoint can also take a `VerifiablePresentation`. Do you need this in this client?

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
```diff
@@ -181,6 +182,41 @@ public CompletionStage<AccessGrant> issue(final URI type, final URI agent, final
...
+                .POST(Request.BodyPublishers.ofByteArray(serialize(presentation))).build();
+
+            return client.send(req, Response.BodyHandlers.ofInputStream())
+                .thenApply(res -> {
+                    try (final InputStream input = res.body()) {
+                        final int status = res.statusCode();
+                        if (isSuccess(status)) {
+                            return processVerificationResult(input);
+                        }
+                        throw new AccessGrantException("Unable to perform Access Grant verify: HTTP error " + status,
```
jholleran, line 210: In the case of an error the response will also contain an `error`, `errorDescription` and an optional `reason`. These could be useful for a downstream client to know but they are getting dropped here and we are only giving them the `status`.

acoburn: According to the VC-API spec, the verify endpoint will return a 200 in cases where the credential is invalid, but there is no defined response for, say, 4xx or 5xx category errors.

I think it would be great for the client to _try_ to parse 4xx/5xx error bodies as if they contain those response, but that can be part of an overall improvement in how errors are processed.

acoburn (approved):
LGTM

jholleran (approved):

PR #404, opened 2023-04-06
JCL-306: added example for working with nested Solid resources

timea-solid:
In the site documentation, I have added an example for the use case
"I would like to model each book as a separate resource so that I can control access on a per book basis. And then I want to link the books with my library which is also a Solid resource,"

This can be then used in the docs.inrupt.com

### DIFF
--- a/src/site/apt/data-modeling/multiple-resources.apt.vm
+++ b/src/site/apt/data-modeling/multiple-resources.apt.vm
@@ -1,0 +1,46 @@
+Data Modeling for multiple Solid Resources
+
+* How to complete
+
+    The solution involves working with domain-specific type such as <<<SolidContainer>>> and <<<SolidResource>>>. 
+
+* Working with Solid Containers
+
+    We have to access the contained resources of a Solid Container. We assume that all Solid RDF Resources are Books. 
+
+    You can use a wrapper class (describe on the {{{./one-resource.html}One Solid Resource}} page) with the high-level Solid client, mapping data seamlessly between your Java application and an RDF resource.
+
++---
+import com.inrupt.client.auth.Session;
+import com.inrupt.client.openid.OpenIdSession;
+import com.inrupt.client.solid.SolidContainer;
+import java.net.URI;
+
+public class MyApplication {
+
+    public void run() {
+        Session session = OpenIdSession.ofClientCredentials(issuer, clientId, clientSecret, "client_secret_basic");
+        SolidSyncClient client = SolidSyncClient.getClient().session(session);
+
+        URI uri = URI.create("https://storage.example/booklibrary/");
+        try (final SolidContainer container = client.read(uri, SolidContainer.class)) {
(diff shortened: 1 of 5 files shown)


src/site/apt/data-modeling/multiple-resources.apt.vm
```diff
@@ -0,0 +1,51 @@
...
+public class MyApplication {
+
+    public void run() {
+        Session session = OpenIdSession.ofClientCredentials(issuer, clientId, clientSecret, "client_secret_basic");
+        SolidSyncClient client = SolidSyncClient.getClient().session(session);
+
+        URI uri = URI.create("https://storage.example/booklibrary/");
+        try (final SolidContainer container = client.read(uri, SolidContainer.class)) {
+            final var resources = container.getResources();
+            resources.stream().forEach(r -> {
```
acoburn, line 28:
```suggestion
            resources.forEach(r -> {
```

src/site/apt/data-modeling/multiple-resources.apt.vm
```diff
@@ -0,0 +1,51 @@
...
+                if (!(r.getIdentifier().toString().endsWith("/"))) { //if it ends in a '/' it is a container
+                    final Request req = Request.newBuilder(r.getIdentifier())
+                        .HEAD(Request.BodyPublishers.noBody())
+                        .build();
+                    if (req.headers().firstValue("Content-Type").get().equalsIgnoreCase("text/turtle")) {
```
acoburn, line 33: It's possible for this Optional to be empty and the server may return "text/turtle; charset=utf8". (There are potentially even better ways to parse this -- split on `;`, use first array item, trim that value and then perform the case insensitive check)


```suggestion
                    if (req.headers().firstValue("Content-Type").filter(h -> h.contains("text/turtle")).isPresent()) {
```

src/site/apt/data-modeling/one-resource.apt.vm
```diff
@@ -0,0 +1,163 @@
...
+        Set<URI> getSongs() {
+            return objects(EX_SONG, TermMappings::asIri, ValueMappings::iriAsUri);
+        }
+
+    }
+}
++---
+
+    Your inner class (<<<Node>>> in this example) will have access to a variety of functions
+    to aid in mapping between the underlying RDF Graph and your Java type. {{{https://javadoc.io/doc/com.inrupt/inrupt-rdf-wrapping-commons/latest/com/inrupt/rdf/wrapping/commons/WrapperBlankNodeOrIRI.html}These methods}} are summarized in the <<<inrupt-rdf-wrapping-commons>>> Javadocs.
```
acoburn, line 86: I would point these URLs to our docs at https://docs.inrupt.com/developer-tools/api/java/inrupt-client/latest/com/inrupt/rdf/wrapping/commons/package-summary.html

acoburn (approved):

PR #406, opened 2023-04-06
JCL-319: fixed examples & corrected bug in doc

timea-solid:
I ran the examples to make sure they work. All good.

### DIFF
--- a/examples/cli/pom.xml
+++ b/examples/cli/pom.xml
@@ -16,8 +16,8 @@
   </description>
 
   <properties>
-    <maven.compiler.target>17</maven.compiler.target>
-    <maven.compiler.source>17</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
+    <maven.compiler.source>11</maven.compiler.source>
 
     <quarkus.version>2.16.5.Final</quarkus.version>
   </properties>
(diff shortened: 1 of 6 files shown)


timea-solid:
Remark:
* to distinguish between Resource types one needs to look at the content-type -> this needs a HEAD call. 
An idea to improve this would be to do the HEAD call already in the [SolidContainer getResources()](https://github.com/inrupt/solid-client-java/blob/1431b9d5082aad8d29fba2d5b33e3cb2d9f0ffb6/solid/src/main/java/com/inrupt/client/solid/SolidContainer.java#L60).

examples/cli/pom.xml
```diff
@@ -16,8 +16,8 @@
   </description>
 
   <properties>
-    <maven.compiler.target>17</maven.compiler.target>
-    <maven.compiler.source>17</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
```
acoburn, line 19: Quarkus generally recommends using Java 17. Is there a reason to move this back to 11?

acoburn (approved):

examples/cli/pom.xml
timea-solid: Thank you for catching this!

PR #410, opened 2023-04-07
JCL-279: Cache abstraction with Guava and Caffeine implementations

acoburn:
This adds a cache abstraction to the client libraries along with two implementations.

The new interfaces are:

* `com.inrupt.client.ClientCache` - A generalizable cache abstraction with `::get`, `::put` and `::invalidate` methods
* `com.inrupt.client.spi.CacheBuilderService` - A generalizable cache builder with a method defining the size and retention duration

There are two new modules that implement these interfaces:

* Caffeine (`inrupt-client-caffeine`) - this implementation is reportedly faster than guava and has some nice features that we will want to use later
* Guava (`inrupt-client-guava`) - this implementation is widely used and will also work on Android

There is also a no-op implementation for cases where no cache support is desired. The no-op implementation is the default if no other service is registered, which also gives the code backwards compatibility.

This code is integrated into the OpenID, UMA and AccessGrant modules for handling metadata caching.

### DIFF
--- a/access-grant/pom.xml
+++ b/access-grant/pom.xml
@@ -57,6 +57,12 @@
     </dependency>
     <dependency>
       <groupId>com.inrupt.client</groupId>
+      <artifactId>inrupt-client-caffeine</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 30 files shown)


timea-solid (approved):
Interesting. 
Elegant implementation.

PR #411, opened 2023-04-07
JCL-328: Deprecate UmaSession

acoburn:
While the `UmaAuthenticationProvider` is used extensively, the `UmaSession` is not really used at all.

Initially, the thought was that the `UmaSession` object would be used to wrap other session objects to make them usable in the context of UMA, but that's not how these APIs evolved. Now, the `UmaSession` is entirely superfluous.


timea-solid (approved):
Very good find! I like it a lot that we keep the code clean and efficient.

PR #412, opened 2023-04-07
JCL-286: Gradle integration

acoburn:
This adds the infrastructure for gradle integration.

Once this is in place, we can add gradle-specific projects, such as android examples. The focus of this PR is only on the structure required for this to be possible.


timea-solid (approved):
I am really new to gradle. I find it curious that is needs to be an own module.

acoburn:
> I find it curious that is needs to be an own module.

Technically, it doesn't need to be a separate module. It could be a parallel system, completely disjoint from the maven build. My concern about keeping it entirely separate, though, is that the two systems (maven and gradle) could get out of sync, which would defeat the purpose of moving gradle-based examples into this repo. By putting this in a maven module, one can run `./mvnw verify` and the gradle portion _also_ runs. Keeping them separate would require remembering to run _both_ `./mvnw verify` and `./gradlew check`

PR #416, opened 2023-04-10
JCL-331: Convenience modules for integrators

acoburn:
While a modular codebase is very useful and flexible, it also makes it harder for someone who is new to the system to know where to begin.

This PR introduces a new artifact to make this easier:

* `inrupt-client-runtime` -- an application developer can include this dependency in an application, and all of the recommended runtime libraries will be included.

~~* `inrupt-client-minimal` -- a library developer can include this dependency in their code. It is just like `inrupt-client-all` except it does not include the runtime modules.~~

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -90,6 +90,11 @@
       </dependency>
       <dependency>
         <groupId>com.inrupt.client</groupId>
+        <artifactId>inrupt-client-runtime</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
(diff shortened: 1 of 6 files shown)


acoburn:
Some additional conversation about this suggests dropping the `inrupt-client-minimal` artifact and changing `*-all` to one of:

* `inrupt-client-default`
* `inrupt-client-runtime`
* `inrupt-client-libs`

timea-solid (approved):

timea-solid (approved):

PR #422, opened 2023-04-14
JCL-335: issue request builder & grant approval calls

timea-solid:
The names are at the moment taken form the JS library:
* Request issueAccessRequest
* CompletionStage<AccessGrant> approveAccessRequest
We can of course rethink the names. 

The issueAccessRequest has a toCompletableFuture().join() now, not sure.


access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
```diff
@@ -200,6 +200,63 @@ public CompletionStage<AccessGrant> issue(final URI type, final URI agent, final
...
+        }).toCompletableFuture().join();
+    }
+
+    /**
+     * Approves an access grant or request.
+     *
+     * @param request an AccessRequest
+     * @return the next stage of completion containing the resulting credential
+     */
+    public CompletionStage<AccessGrant> approveAccessRequest(final AccessRequest request) {
```
acoburn, line 236: I would call this `issueAccessGrant`

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessRequest.java
```diff
@@ -0,0 +1,45 @@
...
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package com.inrupt.client.accessgrant;
+
+import java.util.Map;
+
+/**
+ * An Access Grant issue request.
+ */
+public class AccessRequest {
```
acoburn, line 28: I would have this extend or implement an `AccessCredential` type

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessRequest.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * The grant data.
+     *
+     * @return the grant returned in a serializable format
+     */
+    public final Map<String, Object> data() {
+        return this.data;
+    }
+
+    public AccessRequest(final Map<String, Object> data) {
```
acoburn, line 41: I would make this more structured. A map is too general

acoburn (commented):
Might be worth discussing this in more detail

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
timea-solid: issueAccessRequest name is to specify it is a request and not the actual issue call. 
I call it better issueGrantRequest

timea-solid:
I know we discussed removing the extra class but that would require me to pass Map<String, Object> around. Because of this, I try here to create more architecture.
Is it going in the right direction?

acoburn:
Superseded by #461

PR #432, opened 2023-04-24
JCL 339: Upgrade JsonB module to use Jakarta EE 10

bncummings:
I had to change `<jakarta.json.version>` from 1.1.6 to 2.0.1 (which was the most recent and popular version that didn't cause conflicts). This wasn't explicitly mentioned in the ticket but the package `jakarta.json.JsonException` wasn't recognised in 1.1.6. Would appreciate any feedback as I'm unsure which version would be most appropriate.


acoburn (approved):
LGTM -- I moved the JSON-API dependency to 2.1.1 (newer) and separated that version management from the glassfish implementation dependency

PR #437, opened 2023-05-05
JCL-337: Serialize expiration date as an ISO 8601 string

acoburn:
At present, the access grant library uses the default serialization of `java.time.Instant`, which results in a `long`, which is not supported by a VC-API endpoint. Instead, the value needs to be serialized as an ISO-8601 string.

In addition, the revocation method uses the status list identifier rather than the access grant identifier, which also causes an error.


timea-solid (approved):
I confirm this is working.
Fixed the failing e2e tests too.

PR #446, opened 2023-05-08
JCL-342: Use correct data subject property for access requests

acoburn:
The AccessRequest data model uses the property `isConsentForDataSubject`, as described in the [documentation](https://docs.inrupt.com/ess/latest/services/service-access-grant-issuer/#access-request).

This changes the generation of an access request to follow that structure.


timea-solid (approved):

timea-solid:
If I follow correctly, only the Solid Access Request type of grant has this change.

acoburn:
> If I follow correctly, only the Solid Access Request type of grant has this change.

This is correct. The Access Grant type uses the existing properties

PR #447, opened 2023-05-10
JCL-337: e2e tests improvements

timea-solid:
A few fixes:
* code that changes an ACP + updated documentation
* improved tests based of NonRDf resources & live servers test
* small improvements in auth scenarios and domain level scenarios


integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
```diff
@@ -581,6 +613,69 @@ void accessGrantCreateNonRdfTest(final Session session) throws IOException {
...
+        final IRI acpAccessControl = rdf.createIRI(ACP.accessControl.toString());
+        final IRI aclRead = rdf.createIRI(ACL.Read.toString());
+        final IRI aclWrite = rdf.createIRI(ACL.Write.toString());
+
+        // find the acl Link in the header of the resource
+        final Request req = Request.newBuilder(resourceURI)
+                .HEAD()
+                .build();
+        final Response<Void> res = authClient.send(req, Response.BodyHandlers.discarding());
+        final Headers.Link acrLink = res.headers().allValues("Link").stream()
```
acoburn, line 631: I would have thought that our [Metadata](https://docs.inrupt.com/developer-tools/api/java/inrupt-client/latest/com/inrupt/client/solid/Metadata.html) class already had a mechanism for extracting this URL, but it appears that it doesn't. That would be a nice feature of the `Metadata` class.

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
```diff
@@ -581,6 +613,69 @@ void accessGrantCreateNonRdfTest(final Session session) throws IOException {
...
+            .orElse(null);
+
+        // add the triples needed for access grant
+        if (acrLink != null) {
+            final URI resourceACRurl = acrLink.getUri();
+            final IRI resourceACRiri = rdf.createIRI(resourceACRurl.toString());
+
+            final Dataset dataset = rdf.createDataset();
+            //read the existing triples and add them to the dataset
+            try (final SolidRDFSource resource = authClient.read(resourceACRurl, SolidRDFSource.class)) {
```
acoburn, line 644: Note that `SolidRDFSource` implements `Dataset`, so you could omit the `rdf.createDataset()` above and just call `resource.add(...)`. Then, you can just call `authClient.update(resource);`

src/site/apt/sessions/session-access-grants.apt.vm
```diff
@@ -81,50 +80,50 @@ Request req = Request.newBuilder(sharedTextFileURI)
...
 // add the triples needed for access grant
-if (!acp.isEmpty()) {
-    URI oneAcp = acp.get(0);
-    IRI acpResource = rdf.createIRI(oneAcp.toString());
-    Dataset dataset = rdf.createDataset();
+if (acrLink != null) {
+    URI resourceACRurl = acrLink.getUri();
+    IRI resourceACRiri = rdf.createIRI(resourceACRurl.toString());
+
+    final Dataset dataset = rdf.createDataset();
```
acoburn, line 94: No need to create a new `Dataset`. You can just use the `SolidRDFSource` below.

acoburn (approved):

PR #457, opened 2023-05-15
JCL-350: Better use of GH Environments for CI/CD workflows

acoburn:
GH includes the use of "Environments" for use with CI/CD workflows. These environments are intended to be used as deployment targets, but we are currently using this as a way to partition configuration secrets. This adjusts the CI/CD structure to align our use of environments with the intended use, namely deployment targets.

*Note*: Before merging, this will require adjusting the branch protection rules

### DIFF
--- a/.github/workflows/cd-config.yml
+++ b/.github/workflows/cd-config.yml
@@ -11,8 +11,10 @@ on:
 
 jobs:
   publish:
-    name: Build and deploy
+    name: Deploy artifacts
     runs-on: ubuntu-latest
+    environment:
+      name: "Sonatype"
 
     steps:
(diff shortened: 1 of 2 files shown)


timea-solid (commented):
Observations I see:
- pod or dev the CI runs both java 11 and 17
- dev CIs will be gating now

acoburn:
> Observations I see:
> 
> * pod or dev the CI runs both java 11 and 17
> * dev CIs will be gating now

If you look closely at the condition statement, dev only runs with java 11 and it isn't gating (failures are tolerated)

timea-solid (approved):

PR #458, opened 2023-05-16
JCL-350: Adjust deployment settings

acoburn:
This includes some additional refinement of the CD jobs to align the environment names with names that are used in Jira, so that it is easier to map these events to standard reports.

In brief: development deployments are now called "Development" and production deployments are called "Production" (Documentation deployments are called "Documentation").

The matrix configuration makes it possible to control for "Production" deployments being triggered by a matching tag and "Development" deployments being triggered by merges to `main`.

### DIFF
--- a/.github/workflows/cd-config.yml
+++ b/.github/workflows/cd-config.yml
@@ -10,11 +10,21 @@ on:
       - inrupt-client-[0-9]+.[0-9]+.[0-9]+.Beta[0-9]+
 
 jobs:
-  publish:
+  deployment:
     name: Deploy artifacts
     runs-on: ubuntu-latest
     environment:
-      name: "Sonatype"
+      name: ${{ matrix.envName }}
+    strategy:
+      matrix:
+        envName: ["Development", "Production"]
+        release:
+          - ${{ contains(github.ref, 'inrupt-client-') }}
+        exclude:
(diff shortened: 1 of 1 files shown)


timea-solid (commented):
I am not sure we should even do Githut Pages anymore. I see how naming it Documentation might be misunderstood to the docs.inrupt.com

acoburn:
> I am not sure we should even do Githut Pages anymore. I see how naming it Documentation might be misunderstood to the docs.inrupt.com

I'd be fine with dropping the Maven site documentation. The generation of Javadocs is something I find useful for snapshots; otherwise, we don't see the API docs until they have been released (unless you manually generate them locally)

timea-solid:
> > I am not sure we should even do Githut Pages anymore. I see how naming it Documentation might be misunderstood to the docs.inrupt.com
> 
> I'd be fine with dropping the Maven site documentation. The generation of Javadocs is something I find useful for snapshots; otherwise, we don't see the API docs until they have been released (unless you manually generate them locally)

I also find them useful but we are not linking to them anywhere. I am fine either way, keep or not keep in CD -> I just don't want anyone to expect those to be the reality of the final docs.

timea-solid (approved):

PR #459, opened 2023-05-17
JCL-279: Implement token cache for session objects

acoburn:
This makes use of the new `ClientCache<K, V>` mechanism, applying it to session credential objects.

This required inverting one section of the `ReactiveAuthorization` class. Before, the `Authenticator::authenticate` method accepted a `Session` parameter. Now, the `Session::authenticate` method accepts an `Authenticator` parameter:

https://github.com/inrupt/solid-client-java/blob/17d799b84ec05ac887d46dc815c6f6a411a398ee/api/src/main/java/com/inrupt/client/auth/ReactiveAuthorization.java#L98

The previously used method is now deprecated.

This also changes the way the `OpenIdSession::fromCache` works. Previously, _any_ `fromCache` requests returned an OpenId credential, which incorrectly bypasses reactive authorization. In this case, the is still only a single credential, but the internal cache keeps track of which URIs have performed reactive authorization and (provided that the credential is still valid), that ID Token credential will be provided.


core/src/test/java/com/inrupt/client/core/MockHttpService.java
```diff
@@ -133,7 +133,7 @@ private void setupMocks() {
                     .withHeader("User-Agent", equalTo(USER_AGENT))
                     .withRequestBody(matching("Test String 1"))
                     .withHeader(CONTENT_TYPE, containing(TEXT_PLAIN))
-                    .withHeader("Authorization", containing("Bearer token-67890"))
+                    .withHeader("Authorization", containing("Bearer eyJ"))
```
NSeydoux, line 136: For my own understanding, why was this required?

NSeydoux (approved):

openid/src/main/java/com/inrupt/client/openid/OpenIdSession.java
```diff
@@ -75,16 +79,19 @@ public final class OpenIdSession implements Session {
...
             final Supplier<CompletionStage<Credential>> authenticator) {
         this.id = Objects.requireNonNull(id, "Session id may not be null!");
         this.authenticator = Objects.requireNonNull(authenticator, "OpenID authenticator may not be null!");
         this.dpop = Objects.requireNonNull(dpop);
+        this.requestCache = ServiceProvider.getCacheBuilder().build(1000, Duration.ofMinutes(5));
 
         // Support case-insensitive lookups
         final Set<String> schemeNames = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
         schemeNames.add("Bearer");
+        schemeNames.add("DPoP");
```
hellikopter, line 94: Is it worth using constants from `OpenIdAuthenticationProvider` instead?

openid/src/test/java/com/inrupt/client/openid/OpenIdSessionTest.java
```diff
@@ -157,7 +157,7 @@ void testClientCredentials() {
         assertFalse(session.fromCache(null).isPresent());
         final Optional<URI> principal = session.getPrincipal();
         assertEquals(Optional.of(URI.create(WEBID)), principal);
-        assertTrue(session.fromCache(null).isPresent());
+        assertFalse(session.fromCache(null).isPresent());
```
hellikopter, line 160: Duplicates assertion from 3 lines above

acoburn: In the past, the cache would be populated at this point (which was incorrect). This assertion tests that the cache is still not populated.

openid/src/main/java/com/inrupt/client/openid/OpenIdSession.java
acoburn: Possibly -- I've generally been moving away from storing string constants in a single file and using them elsewhere. It makes no difference to the JVM, since there will only ever be a single instance of the string in memory.

core/src/test/java/com/inrupt/client/core/MockHttpService.java
acoburn: This allows us to vary the mock server response based on the presence of an auth token.

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantSession.java
```diff
@@ -129,7 +165,10 @@ public CompletionStage<Optional<Credential>> authenticate(final Request request,
 
     @Override
     public Optional<Credential> fromCache(final Request request) {
-        // TODO add cache
+        final Credential cachedToken = tokenCache.get(request.uri());
+        if (cachedToken != null && cachedToken.getExpiration().isAfter(Instant.now())) {
```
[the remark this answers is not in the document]
acoburn: I considered this, but I was concerned that there would be a race condition:

If multiple threads are reading/updating the cache, one thread could have fetched the token from the cache (line 168) and before line 169 another thread could have updated the cache. So if the first thread then invalidates the cache, it would have done so unnecessarily, as it would have already been updated by another thread.

hellikopter (approved):

openid/src/main/java/com/inrupt/client/openid/OpenIdAuthenticationProvider.java
```diff
@@ -59,15 +64,20 @@ public String getScheme() {
...
     public Authenticator getAuthenticator(final Challenge challenge) {
         validate(challenge);
         return new OpenIdAuthenticator(priorityLevel);
     }
 
-    static void validate(final Challenge challenge) {
+    void validate(final Challenge challenge) {
         if (challenge == null ||
-                !BEARER.equalsIgnoreCase(challenge.getScheme())) {
+                !schemes.contains(challenge.getScheme())) {
```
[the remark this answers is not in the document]
acoburn: Actually, it's not case sensitive: https://github.com/inrupt/solid-client-java/blob/2ac4a8882c9d81e4ce4c4f465aebaf9dc215547a/openid/src/main/java/com/inrupt/client/openid/OpenIdAuthenticationProvider.java#L45

PR #464, opened 2023-05-19
JCL-312: Refactor future handling

acoburn:
There are many possible approaches to refactoring the use of futures in the SolidSyncClient, but I opted to keep this as simple as possible. If a user wants full control over this there are many alternatives, such as:

* Use the async client
* Configure an HTTP Client with specific timeouts as part of the client construction process
* Use the low-level client and set timeouts on a per-request basis

The high-level client is intended to be simple to use, and this implementation retains that level of simplicity.


langsamu (approved):
The change looks fine. I read it as the unwrapping of an async container exception.

But I don't understand how it relates to the PR description.

acoburn:
@langsamu the PR description here has more to do with the JIRA issue description

PR #473, opened 2023-05-22
JCL-361: Make it easier to modify Access Control resources

acoburn:
This introduces a utility function to make it easier to modify Access Control resources to support AccessGrants.

Eventually, this will become part of a stand-alone ACP module, but adding full support for ACP is beyond the scope of the 1.0 release. This, however, allows developers to interact more easily with ACRs within the 1.0 release timeframe.

A sample usage of this would be:

```java
try (final SolidRDFSource resource = client.read(uri, SolidRDFSource.class)) {
    resource.getMetadata().getAcl().ifPresent(acl -> {
        try (final SolidRDFSource acr = client.read(acl, SolidRDFSource.class)) {
            AccessGrantUtils.accessControlPolicyTriples(acl, ACL.Read, ACL.Write)
                .forEach(acr.getGraph()::add);
            client.update(acr);
        }
    });
}
```

### DIFF
--- a/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
+++ b/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
@@ -391,13 +391,13 @@ public class AccessGrantClient {
         final URI type;
         final Set<String> supportedTypes;
         if (AccessGrant.class.isAssignableFrom(clazz)) {
-            type = ACCESS_GRANT;
+            type = URI.create("SolidAccessGrant");
             supportedTypes = ACCESS_GRANT_TYPES;
         } else if (AccessRequest.class.isAssignableFrom(clazz)) {
-            type = ACCESS_REQUEST;
+            type = URI.create("SolidAccessRequest");
             supportedTypes = ACCESS_REQUEST_TYPES;
         } else if (AccessDenial.class.isAssignableFrom(clazz)) {
-            type = ACCESS_DENIAL;
+            type = URI.create("SolidAccessDenial");
             supportedTypes = ACCESS_DENIAL_TYPES;
         } else {
             throw new AccessGrantException("Unsupported type " + clazz + " in query request");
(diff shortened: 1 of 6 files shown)


access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
```diff
@@ -391,13 +391,13 @@ public <T extends AccessCredential> CompletionStage<List<T>> query(final URI age
         final URI type;
         final Set<String> supportedTypes;
         if (AccessGrant.class.isAssignableFrom(clazz)) {
-            type = ACCESS_GRANT;
+            type = URI.create("SolidAccessGrant");
```
NSeydoux, line 394: For my own understanding, in this case where the given string isn't a URI, what does the constructor do?

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantUtils.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        // Access Control
+        final IRI accessControl = asIRI(URIBuilder.newBuilder(acl).fragment(UUID.randomUUID().toString()).build());
+        triples.add(rdf.createTriple(accessControl, a, asIRI(ACP.AccessControl)));
+        triples.add(rdf.createTriple(accessControl, asIRI(ACP.apply), policy));
+
+        // Access Control Resource
+        final IRI subject = asIRI(acl);
+        triples.add(rdf.createTriple(subject, a, asIRI(ACP.AccessControlResource)));
```
NSeydoux, line 74: Is this triple necessary? I assumed this type would be present by default in the ACR.

NSeydoux (approved):

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
acoburn: This is a relative URI. Perfectly valid

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantUtils.java
acoburn: In the Inrupt implementation, none of the type triples are strictly necessary. That said, a different impl may require them. Furthermore, the ACR for a resource may be empty, in which case the triple will not already be present.

In short: it never hurts to be explicit, and there may be cases where it is necessary

PR #476, opened 2023-05-23
JCL-359: Add denyAccess method to AccessGrantClient

acoburn:
This adds a new method: `AccessGrantClient.denyAccess(AccessRequest)`, allowing a developer to create an access denial receipt.

### DIFF
--- a/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessDenial.java
+++ b/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessDenial.java
@@ -43,7 +43,7 @@ import java.util.stream.Collectors;
 import org.apache.commons.io.IOUtils;
 
 /**
- * An Access Grant abstraction, for use with interacting with Solid resources.
+ * An Access Denial abstraction, for use when interacting with Solid resources.
  */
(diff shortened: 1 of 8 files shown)


NSeydoux (approved):

langsamu (changes requested):

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessDenial.java
```diff
@@ -0,0 +1,214 @@
...
+    private final URI creator;
+    private final Instant expiration;
+    private final Status status;
+
+    /**
+     * Read a verifiable presentation as an AccessDenial.
+     *
+     * @param grant the Access Denial serialized as a verifiable presentation
+     */
+    protected AccessDenial(final String grant) throws IOException {
```
langsamu, line 72: `grant` local here sounds confusing. Is this a request that's being denied? Or is it a denial that's being wrapped?

acoburn: I'll adjust that parameter name

langsamu (approved):

PR #477, opened 2023-05-23
JCL-363: Rework verification response object

acoburn:
The current `VerificationResponse` object is an internal class (POJO) in `AccessCredentialClient`, but it should be more properly structured. This PR moves the simple POJO into its own immutable class. There is still an internal POJO for JSON marshalling.

### DIFF
--- a/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessCredentialVerification.java
+++ b/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessCredentialVerification.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2023 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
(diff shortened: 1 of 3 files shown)


access-grant/src/main/java/com/inrupt/client/accessgrant/AccessCredentialVerification.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * The verification checks that were performed.
+     *
+     * @return a list of any verification checks performed, never {@code null}
+     */
+    public List<String> getChecks() {
+        return checks;
+    }
```
NSeydoux, line 56: Is there any way to make it explicit in the typing that these are immutable?

NSeydoux (approved):

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessCredentialVerification.java
acoburn: Not in the typing, but I can amend the javadocs

PR #491, opened 2023-05-26
JCL-368: Use URIs for purpose arguments

acoburn:
The current JCL API identifies purposes as Strings, but the ESS documentation indicates that purposes are URIs. This aligns the JCL API with the expectation that purposes are always URIs.

For methods that were present and deprecated in earlier releases, those still use String values. All other methods now use `Set<URI>`.

### DIFF
--- a/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
+++ b/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
@@ -181,7 +181,7 @@ public class AccessGrantClient {
      * @return the next stage of completion containing the resulting access request
      */
     public CompletionStage<AccessRequest> requestAccess(final URI agent, final Set<URI> resources,
-            final Set<String> modes, final Set<String> purposes, final Instant expiration) {
+            final Set<String> modes, final Set<URI> purposes, final Instant expiration) {
         Objects.requireNonNull(resources, "Resources may not be null!");
         Objects.requireNonNull(modes, "Access modes may not be null!");
         return v1Metadata().thenCompose(metadata -> {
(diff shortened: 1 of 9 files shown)


access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
```diff
@@ -290,12 +290,20 @@ public CompletionStage<AccessGrant> issue(final URI type, final URI agent, final
         Objects.requireNonNull(type, "Access Grant type may not be null!");
         Objects.requireNonNull(resources, "Resources may not be null!");
         Objects.requireNonNull(modes, "Access modes may not be null!");
+        final Set<URI> uriPurposes = new HashSet<>();
+        for (final String p : purposes) {
+            try {
+                uriPurposes.add(URI.create(p));
+            } catch (final IllegalArgumentException ex) {
+                LOGGER.debug("Ignoring non-URI purpose: {}", ex.getMessage());
```
langsamu, line 298: This sounds like a serious change in behaviour.

I think you should definitely document it. Maybe an `@ImplNote`?

langsamu: Actually this all new functionality, no? So not a change.

But definitely document.

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessRequest.java
```diff
@@ -123,7 +123,8 @@ static AccessRequest parse(final String serialization) throws IOException {
                 final Set<String> modes = asSet(consent.get("mode")).orElseGet(Collections::emptySet);
                 final Set<URI> resources = asSet(consent.get("forPersonalData")).orElseGet(Collections::emptySet)
                     .stream().map(URI::create).collect(Collectors.toSet());
-                final Set<String> purposes = asSet(consent.get("forPurpose")).orElseGet(Collections::emptySet);
+                final Set<URI> purposes = asSet(consent.get("forPurpose")).orElseGet(Collections::emptySet)
+                    .stream().map(URI::create).collect(Collectors.toSet());
```
langsamu, line 127: `.map(URI::create)` might fail, no?

langsamu (changes requested):

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessRequest.java
acoburn: Addressed in https://github.com/inrupt/solid-client-java/pull/491/commits/3b79f521c2335cd4078ed54e706b5e9b7a4a81fe

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
acoburn: Addressed in https://github.com/inrupt/solid-client-java/pull/491/commits/3b79f521c2335cd4078ed54e706b5e9b7a4a81fe

langsamu (approved):

PR #503, opened 2023-05-31
JCL-373: Address minor compiler warnings

acoburn:
There are two categories of changes here:

* For type coersions, we are explicit about cases where those are allowed (`@SuppressWarnings("unchecked")`) -- in each case, the compiler already catches errors, but the type system isn't quite smart enough to know that the casts are legal
* Removal of deprecated methods -- some of our code is intentionally using methods that we have marked as deprecated, but in some cases, we simply haven't updated those code paths to use the latest version


edwardsph (approved):
LGTM

PR #505, opened 2023-06-01
JCL-367: Distinguish between creator and recipient in access credential queries

acoburn:
At present, the `AccessGrantClient.query()` method has an `agent` parameter, which refers to the recipient of an access credential, but there is no way to query based on `creator`. This PR introduces an additional parameter so that one can query for creator and/or recipient.

**Note**: this also moves the `resource` URI to the first position in the new `::query` API (to be introduced in Beta3), as this property is anticipated to be `null` less often.

### DIFF
--- a/access-grant/src/test/java/com/inrupt/client/accessgrant/AccessGrantClientTest.java
+++ b/access-grant/src/test/java/com/inrupt/client/accessgrant/AccessGrantClientTest.java
@@ -508,9 +508,8 @@ class AccessGrantClientTest {
         final String token = generateIdToken(claims);
         final AccessGrantClient client = agClient.session(OpenIdSession.ofIdToken(token));
 
-        final List<AccessGrant> grants = client.query(null,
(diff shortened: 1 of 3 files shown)


access-grant/src/test/java/com/inrupt/client/accessgrant/AccessGrantClientTest.java
NSeydoux, line 1: Tests referencing the notion of `Agent` (e.g. `testQueryGrantAgent`) could be renamed to the appropriate role, i.e. `Recipient` or `Creator`.

NSeydoux (approved):

access-grant/src/test/java/com/inrupt/client/accessgrant/AccessGrantClientTest.java
acoburn: That's a very good idea. I'll create a ticket to make that adjustment in the tests

PR #506, opened 2023-06-01
JCL-375: Clarify javadocs for high level client headers

acoburn:
The javadocs were misleading in the description of how headers added to the client builder would be used. This clarifies how the library works in this regard.


edwardsph (approved):
Perfect - absolutely clear

PR #507, opened 2023-06-01
JCL-376: Capture deployment metrics for releases

acoburn:
This improves our use of the Jira integration for capturing deployment metrics for release workflows.

As it was, production deployments (releases) would not be captured by the Jira integration. By explicitly creating deployments and updating the status, we can more effectively capture this data.

### DIFF
--- a/.github/workflows/cd-config.yml
+++ b/.github/workflows/cd-config.yml
@@ -13,8 +13,8 @@ jobs:
   deployment:
     name: Deploy artifacts
     runs-on: ubuntu-latest
-    environment:
-      name: ${{ matrix.envName }}
+    permissions:
+      deployments: write
     strategy:
       matrix:
         envName: ["Development", "Production"]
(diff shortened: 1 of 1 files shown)


NSeydoux:
When running a workflow in a given environment, a deployment created by default, right? If that's the case, I am correct to assume that the issue solved here is to be able to grab this deployment's id in order to set its status? I'm asking, because setting the environment status is doable via the [GitHub CLI](https://docs.github.com/en/rest/deployments/statuses?apiVersion=2022-11-28), and if it was possible to get the current workflow's deployment ID, we could avoid needing additional dependencies. Is it worth looking into, or am I on the wrong tracks?

acoburn:
This is the reference document I was working from. https://github.com/atlassian/github-for-jira/blob/main/docs/deployments.md, so yes, using the GitHub cli directly would also work.

NSeydoux (approved):
If Atlassian recommands this GH action, it's a good starting point. We can revisit to switch to the CLI later if needed.

PR #515, opened 2023-06-08
JCL-384: Optimize request caching in session objects

acoburn:
At present, the caching mechanism compares the full URI in requests, but it should ignore any query parameters or hash fragments.

### DIFF
--- a/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantSession.java
+++ b/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantSession.java
@@ -152,7 +152,7 @@ public final class AccessGrantSession implements Session {
         return authenticator.authenticate(this, request, algorithms)
             .thenApply(credential -> {
                 if (credential != null) {
-                    tokenCache.put(request.uri(), credential);
+                    tokenCache.put(cacheKey(request.uri()), credential);
                 }
                 return Optional.ofNullable(credential);
             });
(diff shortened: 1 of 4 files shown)


NSeydoux (approved):

langsamu (changes requested):
OIDC [allows](https://openid.net/specs/openid-connect-core-1_0.html#Serializations) messages to be serialized as query strings.

Is this restriction on cache keys not too harsh (especially for the `OpenIdSession`)?

acoburn:
> OIDC [allows](https://openid.net/specs/openid-connect-core-1_0.html#Serializations) messages to be serialized as query strings.

OIDC Messages are [defined](https://openid.net/specs/openid-connect-core-1_0.html#Terminology) as requests and responses sent between the RP (client) and OP (OpenID Provider). This PR has nothing to do with those messages: it is related to caching of credential values for requests between an RP and an RS (Resource Server).

langsamu:
> > OIDC [allows](https://openid.net/specs/openid-connect-core-1_0.html#Serializations) messages to be serialized as query strings.
> 
> OIDC Messages are [defined](https://openid.net/specs/openid-connect-core-1_0.html#Terminology) as requests and responses sent between the RP (client) and OP (OpenID Provider). This PR has nothing to do with those messages: it is related to caching of credential values for requests between an RP and an RS (Resource Server).

I stand corrected.

But is the restriction not too harsh still? Resource Servers might well require different authentication depending on query string.

Even Solid, though the `OpenIdSession` is not specific to Solid.

acoburn:
> But is the restriction not too harsh still? Resource Servers might well require different authentication depending on query string.

This is a good point. I've changed the implementation to only remove the fragment URI (if one exists)

langsamu (approved):

PR #517, opened 2023-06-08
JCL-385: Improve PKCE verifier to always produce spec conforming values

acoburn:
Occasionally the CI/CD workflows will fail because the PKCE verifier is less than 43 characters (required by the spec). This can happen if the random number selected is too low, resulting in an encoded byte array that has too few characters.

This is fixed here by ensuring a minimum value for the BigInteger.

Effectively it changes the range from [0 - 2^256) to [2^256 - 2^257) so the generated byte array is always sufficiently long.

Imagine that the random number generator selects the number 250 -- this will be packed into a single octet and the entropy of the validator will be too low. This way, there will always be at least 32 octets.

### DIFF
--- a/openid/src/test/java/com/inrupt/client/openid/PKCETest.java
+++ b/openid/src/test/java/com/inrupt/client/openid/PKCETest.java
@@ -20,7 +20,6 @@
  */
 package com.inrupt.client.openid;
 
-import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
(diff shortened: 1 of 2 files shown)


NSeydoux (approved):

openid/src/test/java/com/inrupt/client/openid/PKCETest.java
```diff
@@ -30,16 +29,16 @@ class PKCETest {
 
     @Test
     void createChallengeTest() {
-        assertTrue(PKCE.createChallenge("🐶🐶🐶", "SHA-256").getBytes(UTF_8).length >= 43);
-        assertTrue(PKCE.createChallenge("🐶🐶🐶", "SHA-256").getBytes(UTF_8).length <= 128);
-        assertTrue(PKCE.createChallenge("", "SHA-256").getBytes(UTF_8).length >= 43);
-        assertTrue(PKCE.createChallenge("", "SHA-256").getBytes(UTF_8).length <= 128);
+        assertTrue(PKCE.createChallenge("🐶🐶🐶", "SHA-256").length() >= 43);
```
NSeydoux, line 32: For my own understanding, was the `.getBytes(UTF_8)` useful in the first place?

acoburn: No, the getBytes test was unnecessary. The spec requires checking the character length of the resulting string, not the size of the byte array. Admittedly, for ASCII strings, the two will be equivalent, but it is unnecessary

PR #527, opened 2023-06-12
JCL-337: Use verify endpoint in tests

NSeydoux:
Call verify to check the validity of the issued grant, and the invalidity of the revoked grant.

### DIFF
--- a/integration/base/src/main/java/com/inrupt/client/integration/base/MockAccessGrantServer.java
+++ b/integration/base/src/main/java/com/inrupt/client/integration/base/MockAccessGrantServer.java
@@ -31,7 +31,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.net.URI;
-import java.util.List;
+import java.util.Arrays;
+import java.util.Collections;
 
 import org.apache.commons.io.IOUtils;
 
@@ -176,9 +177,9 @@ class MockAccessGrantServer {
                 .withHeader(HEADER_AUTHORIZATION, containing(SCHEME_BEARER))
                 .willReturn(jsonResponse(
                         new AccessCredentialVerification(
-                                List.of("issuanceDate", "proof", "expirationDate", "credentialStatus"),
-                                List.of(),
-                                List.of()),
+                                Arrays.asList("issuanceDate", "proof", "expirationDate", "credentialStatus"),
+                                Collections.emptyList(),
+                                Collections.emptyList()),
                         200)
                 ));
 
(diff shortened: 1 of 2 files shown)


integration/base/src/main/java/com/inrupt/client/integration/base/MockAccessGrantServer.java
```diff
@@ -123,16 +140,66 @@ private void setupMocks() {
...
+        wireMockServer.stubFor(post(urlEqualTo(VERIFY))
+                .inScenario(SCENARIO_ACCESS_GRANT)
+                .whenScenarioStateIs(SCENARIO_STATE_ISSUED)
+                .atPriority(1)
+                .withHeader(HEADER_AUTHORIZATION, containing(SCHEME_BEARER))
+                .willReturn(jsonResponse(
+                        new AccessCredentialVerification(
+                                List.of("issuanceDate", "proof", "expirationDate", "credentialStatus"),
+                                List.of(),
+                                List.of()),
```
acoburn, line 181: If we want to support this in a Java8 runtime (we don't currently but we may with Android), it would be better to use `Arrays.asList(...)` and `Collections.emptyList()`

integration/base/src/main/java/com/inrupt/client/integration/base/MockAccessGrantServer.java
```diff
@@ -123,16 +140,66 @@ private void setupMocks() {
...
+                .inScenario(SCENARIO_ACCESS_GRANT)
+                .whenScenarioStateIs(SCENARIO_STATE_REVOKED)
+                .atPriority(1)
+                .withHeader(HEADER_AUTHORIZATION, containing(SCHEME_BEARER))
+                .withRequestBody(containing("providedConsent"))
+                .willReturn(jsonResponse(
+                        new AccessCredentialVerification(
+                                List.of("issuanceDate", "proof", "expirationDate", "credentialStatus"),
+                                List.of(),
+                                List.of("credentialStatus validation has failed: credential has been revoked")),
```
acoburn, line 195: Ditto about list support in Java 8

acoburn (approved):
LGTM, one minor comment that isn't really required to fix now (more of an FYI)

integration/base/src/main/java/com/inrupt/client/integration/base/MockAccessGrantServer.java
NSeydoux: Addressed in https://github.com/inrupt/solid-client-java/pull/527/commits/b051b636e2e11d1d8a21411de32b103088dc59c7

integration/base/src/main/java/com/inrupt/client/integration/base/MockAccessGrantServer.java
NSeydoux: Addressed in https://github.com/inrupt/solid-client-java/pull/527/commits/b051b636e2e11d1d8a21411de32b103088dc59c7

PR #528, opened 2023-06-13
Downgrade OkHttp to be compatible with Quarkus

edwardsph:


acoburn (approved):
We have seen some problems with OkHttp 4.11 at runtime

PR #531, opened 2023-06-14
Test authentication fails with revoked grant

NSeydoux:
This test that fetching with a revoked grant fails.

This currently fails, because the agent running the test has access without requiring the access grant.


NSeydoux:
Superceded by #535

PR #532, opened 2023-06-14
JCL-382: Access Credential Query builder

acoburn:
This adds an `AccessCredentialQuery` class to make it easier to add permutations to the query interface. In particular, this makes it possible to add multiple access modes ("Read", "Write") and purpose identifiers to a query.

Otherwise, one needs to make multiple requests with the client and manage the set operations manually.

This builder looks like this:

```java
var q = AccessCredentialQuery.newBuilder()
        .resource(uri)
        .recipient(agent)
        .purpose(purpose1)
        .purpose(purpose2)
        .mode("Read")
        .mode("Write")
        .build(AccessGrant.class);

client.query(q)
        .toCompletableFuture()
        .join();
```


kay-kim (approved):
works for me!

timea-solid (approved):
Elegant

access-grant/src/test/java/com/inrupt/client/accessgrant/AccessGrantClientTest.java
NSeydoux, line 1: None of the tests cover cases where either multiple modes or purposes are set, would it be worth adding one?

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessCredentialQuery.java
```diff
@@ -0,0 +1,228 @@
...
+        return type;
+    }
+
+    /**
+     * Get the requested resource.
+     *
+     * @return the resource, may be {@code null}
+     */
+    public URI getResource() {
+        return resource;
```
NSeydoux, line 93: Even if it may not be supported by all Access Management Applications, an Access Grant may cover multiple resources. Is it something that should be reflected in the querying capabilities? Or do we want to push this constraint down to the client libraries, in which case we should consider the same for the JC client.

NSeydoux (approved):

PR #535, opened 2023-06-19
JCL-388: separate resource owner from requestor in access grant e2e tests

timea-solid:
This also:
* removes the non recursive test scenario since this feature is not implemented in JCL


access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
```diff
@@ -544,7 +544,9 @@ public CompletionStage<Void> revoke(final AccessCredential credential) {
      *
      * @param credential the access credential
      * @return the next stage of completion
+     * @deprecated as of Beta4
      */
+    @Deprecated
```
acoburn, line 549: Can you move this into its own PR with a dedicated ticket?

integration/base/src/main/java/com/inrupt/client/integration/base/MockAccessGrantServer.java
```diff
@@ -47,7 +47,7 @@ class MockAccessGrantServer {
 
     private static final String SCHEME_BEARER = "Bearer";
 
-    // An identifier to enable statefulness in Wiremock, useful to manage revocation and verification.
+    // An identifier to enable tastefulness in Wiremock, useful to manage revocation and verification.
```
acoburn, line 50: Revert

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
timea-solid: done

timea-solid:
Decided to exclude the :accessGrantLifecycle tests to run on local mocks because the work involved to further mock the communication between vc service mock and UMA server mock is beyond the point of a local run scenario. Remark: when run on live servers (fully distributed) this tests scenario runs correctly (the default on GitHub).

integration/uma/src/test/resources/simplelogger.properties
```diff
@@ -1 +1 @@
-org.slf4j.simpleLogger.log.org.eclipse.jetty=warn
+org.slf4j.simpleLogger.log.org.eclipse.jetty=warn
```
NSeydoux, line 1: Is this intentional?

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
```diff
@@ -86,20 +83,27 @@ public class AccessGrantScenarios {
...
+    protected static String requesterWebidUrl;
+    private static final String MOCK_RESOURCE_OWNER_USERNAME = "someuser";
+    private static final String MOCK_REQUESTER_USERNAME = "requester";
+
+    private static final String REQUESTER_CLIENT_ID = config.getValue("inrupt.test.requester.client-id", String.class);
+    private static final String REQUESTER_CLIENT_SECRET =
+        config.getValue("inrupt.test.requester.client-secret", String.class);
+
+    private static final String RESOURCE_OWNER_CLIENT_ID =
+        config.getValue("inrupt.test.client-id", String.class);
```
NSeydoux, line 95: I would find it less error-prone to align the naming of the environment variable with the change in this PR, and make it `inrupt.test.resource-owner.client-id` (respectively `inrupt.test.resource-owner.client-secret`)

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
```diff
@@ -214,13 +213,13 @@ static void setup() throws IOException {
     @AfterAll
     static void teardown() {
          //cleanup pod
-        session = OpenIdSession.ofClientCredentials(
+        requesterSession = OpenIdSession.ofClientCredentials(
             URI.create(issuer), //Client credentials
-            CLIENT_ID,
-            CLIENT_SECRET,
+            REQUESTER_CLIENT_ID,
+            REQUESTER_CLIENT_SECRET,
```
NSeydoux, line 219: Shouldn't this be the resource owner session? If the access grant has been revoked, the requester will no longer have access.

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
```diff
@@ -236,252 +235,254 @@ static void teardown() {
     @ParameterizedTest
     @MethodSource("provideSessions")
     @DisplayName(":accessGrantLifecycle Access Grant issuance lifecycle")
-    void accessGrantIssuanceLifecycleTest(final Session session) {
+    void accessGrantIssuanceLifecycleTest(final Session resourceOwnerSession, final Session requesterSession) {
+        assumeFalse(ACCESS_GRANT_PROVIDER.contains("localhost"));
```
NSeydoux, line 239: A comment documenting why this is added might be helpful to future us.

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
```diff
@@ -631,74 +653,108 @@ void accessGrantSetNonRdfTest(final Session session) throws IOException {
...
-        // find the acl Link in the header of the resource
-        try (final SolidNonRDFSource resource = authClient.read(resourceURI, SolidNonRDFSource.class)) {
-            resource.getMetadata().getAcl().ifPresent(acl -> {
-                try (final SolidRDFSource acr = authClient.read(acl, SolidRDFSource.class)) {
-                    AccessGrantUtils.accessControlPolicyTriples(acl, ACL.Read, ACL.Write)
-                        .forEach(acr.getGraph()::add);
-                    authClient.update(acr);
-                }
-            });
+    private static <T extends SolidResource> void prepareACPofResource(final SolidSyncClient authClient,
```
NSeydoux, line 709:
```suggestion
    private static <T extends SolidResource> void prepareAcpOfResource(final SolidSyncClient authClient,
```

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
```diff
@@ -236,252 +235,254 @@ static void teardown() {
...
 
-        try (final SolidRDFSource resource = newClient.read(testRDFresourceURI, SolidRDFSource.class)) {
+        try (final SolidRDFSource resource = requesterClient.read(testRDFresourceURI, SolidRDFSource.class)) {
             assertTrue(resource.getMetadata().getContentType().contains(Utils.TEXT_TURTLE));
         }
 
-        newClient.delete(testRDFresourceURI);
-        assertDoesNotThrow(accessGrantClient.revoke(grant).toCompletableFuture()::join);
+        //cleanup
+        resourceOwnerClient.delete(testRDFresourceURI);
```
NSeydoux, line 449: For other resources, this is done in the BeforeAll and AfterAll functions. Should it be the case for this too? This applies to other tests with RDF resources too. In particular, I'm wondering wether the current setup would properly clean up after a failed test.

integration/uma/src/test/resources/simplelogger.properties
timea-solid: might be due to my IDE switch

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
timea-solid: I see. I took a peak at the js ones and they are OWNER and REQUESTER. 
I want to do this in a separate PR because I also want to change the OWNER info used in GitHub.

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
timea-solid: Good catch!

integration/uma/src/test/resources/simplelogger.properties
timea-solid: Ups

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
timea-solid: The resource is created locally reason why we delete it immediately here.

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
```diff
@@ -236,252 +235,256 @@ static void teardown() {
...
-    @ParameterizedTest
-    @MethodSource("provideSessions")
-    @DisplayName(":accessGrantQueryByRequestor Lookup Access Grants by requestor")
-    void accessGrantQueryByRequestorTest(final Session session) {
-        LOGGER.info("Integration Test - Lookup Access Grants by requestor");
+        //query for all grants with a dedicated purpose
+        final AccessCredentialQuery<AccessGrant> query = AccessCredentialQuery.newBuilder()
+                .creator(URI.create(requesterWebidUrl)).build(AccessRequest.class);
+        final List<AccessGrant> grants = resourceOwnerAccessGrantClient.query(query).toCompletableFuture().join();
+        assertTrue(grants.size() >= 1); //on a second run there would be 2 grants because grants are never deleted
```
acoburn, line 358: This seems like a test that could easily hide an underlying problem. It might be better to scope the query to a resource url so that you will never end up with more than one result.

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
```diff
@@ -236,252 +235,256 @@ static void teardown() {
...
 
-        final AccessGrantClient accessGrantClient = new AccessGrantClient(
-                URI.create(ACCESS_GRANT_PROVIDER)
-        ).session(session);
+        //query for all grants of dedicated purpose combinations
```
acoburn, line 360: How is this test checking for dedicated purposes?

timea-solid: ups, left comment when copying test.

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
timea-solid: This is solved now.

integration/uma/src/test/resources/simplelogger.properties
```diff
@@ -1 +1 @@
-org.slf4j.simpleLogger.log.org.eclipse.jetty=warn
+org.slf4j.simpleLogger.log.org.eclipse.jetty=INFO
```
NSeydoux, line 1: Is there a reason for this change?

NSeydoux (approved):

integration/uma/src/test/resources/simplelogger.properties
acoburn: This is useful for debugging, but it would (likely) be best to keep it at `warn` level

PR #483, opened 2023-05-25
JCL-251: Increase test coverage

hellikopter:
Changes in PR:
- expand `maven-surefire-plugin` executions to cover all provided implementations

Implementations

| Interface | Implementation | Module |
|-----------|-------------------|----------|
| ClientCache | CaffeineCache, GuavaCache |  caffeine, guava |
| HttpService | HttpClientService, OkHttpService | httpclient, okhhtp |
| JsonService | JacksonService, JsonbService | jackson, jsonb |
| RdfService | JenaService, RDFLegacyService, RDF4JService | jena, rdf-legacy, rdf4j |

Test coverage

| Module | Dependency implementation matrix |
|----------|-----------------------------------------|
| access-grant | [httpclient, okhhtp] x [jackson, jsonb*] x [caffeine, guava] |
| core | [httpclient, okhhtp] x [jackson, jsonb*] x [jena, rdf-legacy*, rdf4j*] |
| integration | [httpclient*, okhhtp] x [jackson, jsonb*] x [jena, rdf-legacy*, rdf4j*] |
| jena | [httpclient, okhhtp] |
| openid | [httpclient, okhhtp] x [jackson, jsonb*] x [caffeine, guava] |
| rdf-legacy | [httpclient, okhhtp] |
| rdf4j | [httpclient, okhhtp] |
| solid | [httpclient, okhhtp*] x [jena, rdf-legacy*, rdf4j] |
| uma | [httpclient, okhhtp] x [jackson, jsonb*] x [caffeine, guava] |
| webid | [httpclient, okhhtp] x [jena, rdf-legacy*, rdf4j] |

Test variations containing module marked with `*` are ~~disabled~~ removed due to the test failures.

PR #542, opened 2023-06-19
JCL-371: Fix JSON-B config

NSeydoux:
This re-enables the surefire execution configurations disabled in https://github.com/inrupt/solid-client-java/pull/483/commits/d90f442745c35dfe21870f5281b3accafe45c966 relating to the `access-grants` module. Having the tests run required two changes:
- Making a POJO class used for JSON mapping public
- Preventing the JSONB instance to be shared. In some cases, such as queries, where the mocks would serve the same resource several times in parallel, it resulted in parsing errors that appeared to be related to the JSONB shared instance not being thread-safe.

### DIFF
--- a/access-grant/pom.xml
+++ b/access-grant/pom.xml
@@ -195,6 +195,36 @@
               </classpathDependencyExcludes>
             </configuration>
           </execution>
+          <execution>
+            <id>httpclient-jsonb-caffeine-test</id>
+            <phase>test</phase>
+            <goals>
+              <goal>test</goal>
+            </goals>
+            <configuration>
+              <rerunFailingTestsCount>2</rerunFailingTestsCount>
+              <classpathDependencyExcludes>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-okhttp</classpathDependencyExclude>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-jackson</classpathDependencyExclude>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-guava</classpathDependencyExclude>
+              </classpathDependencyExcludes>
+            </configuration>
+          </execution>
+          <execution>
+            <id>httpclient-jsonb-guava-test</id>
+            <phase>test</phase>
+            <goals>
+              <goal>test</goal>
+            </goals>
+            <configuration>
+              <rerunFailingTestsCount>2</rerunFailingTestsCount>
+              <classpathDependencyExcludes>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-okhttp</classpathDependencyExclude>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-jackson</classpathDependencyExclude>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-caffeine</classpathDependencyExclude>
+              </classpathDependencyExcludes>
+            </configuration>
+          </execution>
         </executions>
       </plugin>
       <plugin>


access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
```diff
@@ -922,7 +922,7 @@ static boolean isAccessDenial(final URI type) {
     /**
      * A data object for verification responses.
      */
-    static class VerificationResponseData {
+    static public class VerificationResponseData {
```
acoburn, line 925: Does this need to be public? It was intentionally non-public

jsonb/src/main/java/com/inrupt/client/jsonb/JsonbService.java
```diff
@@ -39,21 +38,21 @@
...
         config.withPropertyNamingStrategy(PropertyNamingStrategy.LOWER_CASE_WITH_DASHES);
-        this.jsonb = JsonbBuilder.create(config);
+        this.config = config;
     }
 
     @Override
     public <T> void toJson(final T object, final OutputStream output) throws IOException {
         try {
-            jsonb.toJson(object, output);
+            JsonbBuilder.create(config).toJson(object, output);
```
acoburn, line 55: Creating a Jsonb object is relatively expensive. Creating it once was intentional

NSeydoux: It appeared to resolve the test issues locally, but that's actually not the case depending on how I run the tests, so I reverted this and I am still investigating.

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
NSeydoux: That's my understanding of the following error: 
```
Caused by: jakarta.json.bind.JsonbException: Error accessing field 'checks' declared in 'class com.inrupt.client.accessgrant.AccessGrantClient$VerificationResponseData'
...
Caused by: java.lang.IllegalAccessException: class is not public: com.inrupt.client.accessgrant.AccessGrantClient$VerificationResponseData.checks/java.util.List/getField, from public Lookup
```
The error is not thrown anymore after the class is made public.

NSeydoux:
This will need to be rebased on https://github.com/inrupt/solid-client-java/pull/543 to clear the dependency check failure.

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
```diff
@@ -430,32 +430,28 @@ private <T extends AccessCredential> CompletionStage<List<T>> query(final URI re
...
-                                        "Unexpected I/O exception while processing Access Grant query", ex);
-                            }
-                        }).toCompletableFuture())
-                .collect(Collectors.toList());
-
-            return CompletableFuture.allOf(futures.toArray(new CompletableFuture[0]))
-                .thenApply(x -> futures.stream().map(CompletableFuture::join).flatMap(List::stream)
-                        .collect(Collectors.toList()));
-        });
+        return v1Metadata().thenApply(metadata -> buildQuery(config.getIssuer(), type,
```
acoburn, line 433: I'm not sure about the refactoring of this code. The use of CompletableFuture.allOf was intended to ensure that multiple requests were performed in parallel without being blocked by any particular response.

I know you mentioned that Jsonb objects are not thread safe, but that contradicts the contract stated in the javadocs:

    All the methods in this class are safe for use by multiple concurrent threads.

https://javaee.github.io/javaee-spec/javadocs/javax/json/bind/Jsonb.html

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
```diff
@@ -925,7 +917,7 @@ static boolean isAccessDenial(final URI type) {
     /**
      * A data object for verification responses.
      */
-    static class VerificationResponseData {
+    public static class VerificationResponseData {
```
acoburn, line 920: If we are going to make this public, then I'd add proper getters and setters to the class

access-grant/src/main/java/com/inrupt/client/accessgrant/Utils.java
```diff
@@ -44,6 +45,8 @@ public static Status asRevocationList2020(final Map<String, Object> credentialSt
                 idx = Integer.parseInt((String) index);
             } else if (index instanceof Integer) {
                 idx = (Integer) index;
+            } else if (index instanceof BigDecimal) {
```
acoburn, line 48: Why would an index value ever be larger than 100,000? As such a BigDecimal would not be needed. Plus, this is for index values, which are always integers. Even if we used very large numbers, a BigInteger would be the right value. Any non-integer value is invalid

acoburn (changes requested):
There are some parts of this PR that don't look quite right to me

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
acoburn: Instead of making this inner class public, I would be more inclined to remove it and make the existing publicly facing class mutable

NSeydoux: Fixed in 23cc8dec835cf9b60062ce5c41bcdfbdcaa9f5ef

access-grant/src/main/java/com/inrupt/client/accessgrant/Utils.java
NSeydoux: The `BigDecimal` class is instantiated by JSONB when parsing an integer from JSON (see the debug screenshot attached). This results in the index being unmodified, and left to its `-1` initialization value, if not checking for `BigDecimal`.

Capture d’écran du 2023-06-21 10-35-48

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessCredentialVerification.java
```diff
@@ -73,6 +82,37 @@ public List<String> getErrors() {
...
 
+    /**
+     * Initialize the verification checks that were performed. This can only be called once, as the checks list is
+     * made unmodifiable.
+     *
+     * @param checks a list of any verification checks performed, never {@code null}
+     */
+    public void setChecks(final List<String> checks) {
+        this.checks = checks;
+        makeImmutable(this.checks);
```
acoburn, line 93:
```suggestion
        this.checks = makeImmutable(checks);
```

All the setters should be adjusted similarly

NSeydoux: Fixed in e426ab78baf5781514d4c762597db99407cfceb3

acoburn (approved):

NSeydoux:
This still introduces flakyness, and the root issue hasn't been resolved, so closing this for now as it isn't a blocker, and we have a ticket to investigate later.

PR #546, opened 2023-06-22
JCL-391: switch env var to new dev

timea-solid:
Also, switch log level from info to warn in base module to be consistent in all modules


timea-solid:
Also switched to another prod webId.

NSeydoux (approved):

PR #548, opened 2023-06-23
JCL-371: Integration tests, rdf4j-legacy

NSeydoux:


NSeydoux:
The integration tests have Jena-specific setup in CoreModulesResource, so excluding Jena will necessarily fail.

PR #549, opened 2023-06-23
JCL-371: Integration tests rdf4j

NSeydoux:


NSeydoux:
The integration tests have Jena-specific setup, so excluding Jena necessarily fails them.

PR #550, opened 2023-06-23
JCL-371: integration module, jsonb dependency

NSeydoux:
This enables running the integration tests with the JSON-B module implementing the JSON service.


jsonb/src/main/java/com/inrupt/client/jsonb/JsonbService.java
```diff
@@ -33,20 +33,28 @@
...
+
     private final Jsonb jsonb;
 
     /**
      * Create a JSON-B service.
      */
     public JsonbService() {
         final JsonbConfig config = new JsonbConfig();
-        config.withPropertyNamingStrategy(PropertyNamingStrategy.LOWER_CASE_WITH_DASHES);
+        config.withPropertyNamingStrategy(new LowerCamelCase());
```
langsamu, line 57: This is a very general change.
_All_ JSON-B functionality from now on willuse this new naming strategy.

I have no problem with it, except to note that we'll want to communicate in release notes.

jsonb/src/main/java/com/inrupt/client/jsonb/JsonbService.java
```diff
@@ -76,5 +84,28 @@ public <T> T fromJson(final InputStream input, final Type type) throws IOExcepti
             throw new IOException("Error parsing JSON", ex);
         }
     }
+
+    static class LowerCamelCase implements PropertyNamingStrategy {
```
langsamu, line 88: Do we not have a 'standard' implementation of this naming strategy?

By standard I mean one that we don't have to implement ourselves.

jsonb/src/main/java/com/inrupt/client/jsonb/JsonbService.java
langsamu: On second thought, I do see a problem.

JSON-B has its default naming strategy (whatever that is).
Now we make our JSON-B facilities use a different one from the default.
And it's actually one we've written ourselves.
This might confuse developers.

At minimum I'd like to see this prominently reflected in javadocs.
Possibly I'd like us to reconsider this.

Perhaps the use of this naming strategy belongs on a level different from _the_ JSON-B class?

langsamu (changes requested):
I know I myself suggested the use of a naming strategy for controlling JSON serialization.

However see some comments about this particular implementation.

PR #557, opened 2023-06-27
JCL-264: improved test, formating & documentation

timea-solid:
Minor improvements of integration tests.

### DIFF
--- a/integration/README.md
+++ b/integration/README.md
@@ -1,6 +1,6 @@
 # Integration tests
 
-The integration tests contain two scenarios of tests:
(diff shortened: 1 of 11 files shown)


langsamu (approved):
Curious, why the addition of the user-agent request header to the mocks?

timea-solid:
> Curious, why the addition of the user-agent request header to the mocks?

Desire to be consistent with other mocks. But than again, it is only locally where one would see these. It is just a mere convenience that could be debated away...

PR #558, opened 2023-06-27
JCL-389: removed copyright year from all headers

timea-solid:
In all java classes and in the license text.


NSeydoux (approved):

timea-solid:
Pushed by mistake to this branch something from JCL-403. I reverted it.

PR #560, opened 2023-06-28
JCL-403: fully automating e2e setup

timea-solid:
We make use of a 'public' container in our tests to execute scenarios with unauthenticated users. For this purpose, we need to change the ACR or this container to allow http://www.w3.org/ns/solid/acp#PublicAgent to read and write from the container.
Also refactored some parts of the code.

### DIFF
--- a/integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
+++ b/integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
@@ -26,6 +26,7 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assumptions.assumeFalse;
 
+import com.inrupt.client.Headers;
 import com.inrupt.client.Request;
 import com.inrupt.client.Response;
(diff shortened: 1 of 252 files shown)


langsamu (approved):

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
```diff
@@ -757,4 +755,25 @@ private static Stream<Arguments> provideSessions() throws SolidClientException {
...
+        final Session session = OpenIdSession.ofClientCredentials(
+                URI.create(issuer), //Client credentials
+                RESOURCE_OWNER_CLIENT_ID,
+                RESOURCE_OWNER_CLIENT_SECRET,
+                AUTH_METHOD);
+
+        authResourceOwnerClient = SolidSyncClient.getClient().session(session);
+    }
+
+    private static void createContainer(final URI publicContainerURI) {
```
NSeydoux, line 769: Should this be considered for the public API of `SolidClient`? It seems like something  a lot of users will need.

timea-solid: good point :) Will make a note about it.

PR #561, opened 2023-06-29
JCL-412: deal with guava 31.1. CVE

timea-solid:
https://nvd.nist.gov/vuln/detail/CVE-2023-2976
I excluded the lib in wiremock where we could. The other modules make use of it in testing so I upgraded it.

### DIFF
--- a/integration/base/pom.xml
+++ b/integration/base/pom.xml
@@ -113,17 +113,20 @@
             <scope>provided</scope>
             <exclusions>
                 <exclusion>
+        <!-- transitive dependency via wiremock -->
+        <!-- when wiremock is updated beyond 2.35, this can be removed -->
                     <groupId>commons-fileupload</groupId>
                     <artifactId>commons-fileupload</artifactId>
                 </exclusion>
-                <exclusion>
-                    <groupId>com.google.guava</groupId>
-                    <artifactId>guava</artifactId>
-                </exclusion>
             </exclusions>
         </dependency>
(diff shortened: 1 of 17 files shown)


integration/base/pom.xml
```diff
@@ -113,13 +113,20 @@
             <scope>provided</scope>
             <exclusions>
                 <exclusion>
+        <!-- transitive dependency via wiremock -->
+        <!-- when wiremock is updated beyond 2.35, this can be removed -->
```
acoburn, line 117: This line refers to json-smart -- can you move it back to right before the json-smart dependency?

jsonb/pom.xml
```diff
@@ -59,6 +59,10 @@
             <groupId>net.minidev</groupId>
             <artifactId>accessors-smart</artifactId>
         </exclusion>
+        <exclusion>
```
acoburn, line 62: Rather than exclude guava, can we just manually manage the version, as we're doing elsewhere?

test/pom.xml
```diff
@@ -43,14 +43,21 @@
       <version>${wiremock.version}</version>
       <scope>provided</scope>
       <exclusions>
+    <!-- transitive dependency via wiremock -->
+    <!-- when wiremock is updated beyond 2.35, this can be removed -->
```
acoburn, line 47: This line is supposed to refer to json-smart, can you move it to that section of the POM?

acoburn (approved):

PR #562, opened 2023-06-30
JCL-401: fetch access requests

timea-solid:
This is adding a more complex test scenarios of the fetch endpoint.
Scenario. 
Alice owns a resource Bob wants to access.
Bob makes an access request
Bob calls the fetch with the id -> this works.
Alice calls the fetch with the same id -> this works too. (because the resource belongs to Alice).

Reworded the java doc on the usage of the recipient in the accessRequest API. (I misunderstood it initially)


acoburn (approved):
LGTM

PR #573, opened 2023-07-03
JCL-394: Issued At javadoc

acoburn:
The issuance date is already supported by AccessCredential subtypes, but the base class is missing a javadoc.


timea-solid:
I saw this part of the issuance date.
But isn't the issue(..) method supposed to have a `final Instant issuanceDate` in its call such that the request, for example, can be requested for a future date?

NSeydoux (approved):

PR #575, opened 2023-07-05
JCL-411: RDF4J service test coverage

NSeydoux:
This adds unit test coverage for the RDF4J implementations of the RDF service. 

The integration tests have a jena-specific part for the core module, which is excluded from the tests running with the RDF4J dependencies, but the domain model mapping integration tests run with all three implementations of the RDF service.

The RDF-specific parts of the core module unit tests are duplicated, so that there is a Jena-specific and an RDF4J-specific implementation of the tests, each being appropriately excluded when running the tests.


core/pom.xml
```diff
@@ -183,13 +183,20 @@
             <id>default-test</id>
             <phase>none</phase>
           </execution>
+          <!-- The rdf-legacy module is an older version of RDF4J specifically targeting Android development. It isn't
```
acoburn, line 186: This explanation is fine, but why are we [including](https://github.com/inrupt/solid-client-java/blob/604de7e4f4931e911e71dc2ca6bfeef8f1905386/core/pom.xml#L150-L155) rdf-legacy in the dependencies in the first place?

core/pom.xml
```diff
@@ -243,6 +259,78 @@
...
+            </goals>
+            <configuration>
+              <excludes>
+                <exclude>**/*JenaTest.java</exclude>
+              </excludes>
+              <classpathDependencyExcludes>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-okhttp</classpathDependencyExclude>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-jsonb</classpathDependencyExclude>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-jena</classpathDependencyExclude>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-rdf-legacy</classpathDependencyExclude>
```
acoburn, line 276: We would not need this special exclusion if we didn't add it to the dependencies in the first place.

core/pom.xml
```diff
@@ -243,6 +259,78 @@
...
+            </goals>
+            <configuration>
+              <excludes>
+                <exclude>**/*JenaTest.java</exclude>
+              </excludes>
+              <classpathDependencyExcludes>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-httpclient</classpathDependencyExclude>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-jsonb</classpathDependencyExclude>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-jena</classpathDependencyExclude>
+                <classpathDependencyExclude>com.inrupt.client:inrupt-client-rdf-legacy</classpathDependencyExclude>
```
acoburn, line 294: ditto

acoburn (commented):
Why not j

NSeydoux:
Let me merge https://github.com/inrupt/solid-client-java/pull/588 into this branch, I initially intended to remove the `rdf-legacy` dependency in its own separate PR, but I don't mind the two changes being squashed into the same commit.

timea-solid (approved):

PR #576, opened 2023-07-05
JCL-419: harden e2e code

timea-solid:
We saw lately too many CI problems. On investigation, it turned out that the problem occurs because the tests do not do a proper cleanup.
A solution here is to also clean up before starting the tests (making sure the containers are empty for example).

It still happens that some tests randomly fail, I am not sure if we can simply up the retry number...

### DIFF
--- a/integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
+++ b/integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
@@ -94,10 +94,6 @@ public class AccessGrantScenarios {
         .orElse("client_secret_basic");
 
     protected static String ACCESS_GRANT_PROVIDER;
-    private static final String PRIVATE_RESOURCE_PATH = config
-        .getOptionalValue("inrupt.test.private-resource-path", String.class)
-        .orElse("private");
-
     protected static final String GRANT_MODE_READ = "Read";
     private static final String GRANT_MODE_APPEND = "Append";
     private static final String GRANT_MODE_WRITE = "Write";
@@ -137,7 +133,6 @@ public class AccessGrantScenarios {
                 .build()
                 .toString());
 
-        State.PRIVATE_RESOURCE_PATH = PRIVATE_RESOURCE_PATH;
         State.WEBID = URI.create(webidUrl);
         final SolidSyncClient client = SolidSyncClient.getClientBuilder().build();
         try (final WebIdProfile profile = client.read(URI.create(webidUrl), WebIdProfile.class)) {
--- a/integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
+++ b/integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
@@ -219,7 +219,7 @@ public class AccessGrantScenarios {
             if (privateContainerURI != null) {
                 Utils.cleanContainerContent(authResourceOwnerClient, privateContainerURI);
             }
-        }  catch (SolidClientException ex) {
+        }  catch (SolidClientException ignored) {
             //do nothing because we are only cleaning up
         }
 
--- a/integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
+++ b/integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
@@ -23,7 +23,6 @@ package com.inrupt.client.integration.base;
 import static org.junit.jupiter.api.Assertions.*;
 import static org.junit.jupiter.api.Assumptions.assumeFalse;
 
-import com.inrupt.client.Headers;
 import com.inrupt.client.Request;
 import com.inrupt.client.Response;
 import com.inrupt.client.accessgrant.*;
@@ -33,7 +32,6 @@ import com.inrupt.client.solid.*;
 import com.inrupt.client.spi.RDFFactory;
 import com.inrupt.client.util.URIBuilder;
 import com.inrupt.client.vocabulary.ACL;
-import com.inrupt.client.vocabulary.LDP;
 import com.inrupt.client.webid.WebIdProfile;
 
 import java.io.ByteArrayInputStream;
@@ -154,17 +152,21 @@ public class AccessGrantScenarios {
             podUrl += Utils.FOLDER_SEPARATOR;
         }
 
-
         privateContainerURI = URIBuilder.newBuilder(URI.create(podUrl))
                 .path(State.PRIVATE_RESOURCE_PATH + "/").build();
+
+        authResourceOwnerClient = createAuthenticatedClient();
+        //if a tests fails it can be that the cleanup was not properly done, so we do it here too
+        Utils.cleanContainerContent(authResourceOwnerClient, privateContainerURI);
+        client.send(Request.newBuilder(privateContainerURI).DELETE().build(), Response.BodyHandlers.discarding());
+        Utils.createContainer(authResourceOwnerClient, privateContainerURI);
 
         testContainerURI = URIBuilder.newBuilder(URI.create(podUrl))
                 .path(State.PRIVATE_RESOURCE_PATH)
                 .path("accessgrant-test-" + UUID.randomUUID() + "/")
                 .build();
 
-        createAuthenticatedClient();
-        createContainer(testContainerURI);
+        Utils.createContainer(authResourceOwnerClient, testContainerURI);
         prepareAcpOfResource(authResourceOwnerClient, testContainerURI, SolidRDFSource.class);
 
         sharedTextFileURI = URIBuilder.newBuilder(testContainerURI)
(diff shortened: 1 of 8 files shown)


integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
```diff
@@ -154,17 +152,19 @@ static void setup() throws IOException {
             podUrl += Utils.FOLDER_SEPARATOR;
         }
 
-
         privateContainerURI = URIBuilder.newBuilder(URI.create(podUrl))
                 .path(State.PRIVATE_RESOURCE_PATH + "/").build();
 
+        authResourceOwnerClient = createAuthenticatedClient();
+        //if a tests fails it can be that the cleanup was not properly done, so we do it here too
+        Utils.deleteContentsRecursively(authResourceOwnerClient, privateContainerURI);
```
acoburn, line 160: We can't do this deletion here, because it will cause any other concurrently-running test to fail.

integration/base/src/main/java/com/inrupt/client/integration/base/AuthenticationScenarios.java
```diff
@@ -125,23 +121,31 @@ static void setup() {
...
+                .path("auth-test-" + UUID.randomUUID())
                 .build();
 
             publicContainerURI = URIBuilder.newBuilder(URI.create(podUrl))
                     .path(PUBLIC_RESOURCE_PATH + "/").build();
 
-            createContainer(publicContainerURI);
-            prepareACR(publicContainerURI);
+            //if a tests fails it can be that the cleanup was not properly done, so we do it here too
+            Utils.deleteContentsRecursively(localAuthClient, publicContainerURI);
```
acoburn, line 147: As mentioned elsewhere, we can't delete this data here, because it will interfere with other concurrently-running tests.

integration/base/src/main/java/com/inrupt/client/integration/base/AuthenticationScenarios.java
```diff
@@ -159,31 +163,28 @@ static void setup() {
 
         privateContainerURI = URIBuilder.newBuilder(URI.create(podUrl))
                 .path(PRIVATE_RESOURCE_PATH + "/").build();
+        //if a tests fails it can be that the cleanup was not properly done, so we do it here too
```
acoburn, line 166: If _tests_ fail, the cleanup (`@AfterAll`) gets run, it's only if the setup (`@BeforeAll`) fails that the cleanup doesn't happen.

integration/base/src/main/java/com/inrupt/client/integration/base/AuthenticationScenarios.java
```diff
@@ -159,31 +163,28 @@ static void setup() {
...
+        Utils.deleteContentsRecursively(localAuthClient, privateContainerURI);
```
acoburn, line 167: We don't want to delete data in any setup phase, as it may interfere with other tests.

integration/base/src/main/java/com/inrupt/client/integration/base/CoreModulesResource.java
```diff
@@ -130,22 +130,29 @@ static void setup() {
...
                 .build();
 
             publicContainerURI = URIBuilder.newBuilder(URI.create(podUrl))
                     .path(PUBLIC_RESOURCE_PATH + "/").build();
-            createAuthenticatedClient();
-            createContainer(publicContainerURI);
-            prepareACR(publicContainerURI);
+
+            //if a tests fails it can be that the cleanup was not properly done, so we do it here too
+            Utils.deleteContentsRecursively(localAuthClient, publicContainerURI);
```
acoburn, line 154: ditto

integration/base/src/main/java/com/inrupt/client/integration/base/DomainModulesResource.java
```diff
@@ -122,22 +121,30 @@ static void setup() {
...
+                .path("domain-test-" + UUID.randomUUID())
                 .path(testContainer)
                 .build();
 
             publicContainerURI = URIBuilder.newBuilder(URI.create(podUrl))
                     .path(PUBLIC_RESOURCE_PATH + FOLDER_SEPARATOR).build();
-            createContainer(publicContainerURI);
-            prepareACR(publicContainerURI);
+            //if a tests fails it can be that the cleanup was not properly done, so we do it here too
+            Utils.deleteContentsRecursively(localAuthClient, publicContainerURI);
```
acoburn, line 146: ditto

integration/base/src/main/java/com/inrupt/client/integration/base/Utils.java
```diff
@@ -166,6 +173,96 @@ public static Set<Triple> publicAgentPolicyTriples(final URI acl) {
         return triples;
     }
 
+    public static void createPublicContainer(final SolidSyncClient authClient, final URI publicContainerURI) {
+        if (!exists(authClient, publicContainerURI)) {
+            try (SolidContainer newContainer = new SolidContainer(publicContainerURI)) {
+                try (final SolidContainer container = authClient.create(newContainer)) {
+                    container.getMetadata().getAcl().ifPresent(acl -> {
+                        int i = 0; // in case we cannot read the acl, we try 2 times before giving up
```
acoburn, line 181: This issue this method is fixing is (I believe) symptomatic of some other issue, and it would be better to fix the underlying issue.

For example, the ACL is not present because another process has just deleted the corresponding resource.

integration/base/src/main/java/com/inrupt/client/integration/base/Utils.java
```diff
@@ -166,6 +173,96 @@ public static Set<Triple> publicAgentPolicyTriples(final URI acl) {
...
+        }
+    }
+
+    static void deleteContentsRecursively(final SolidSyncClient client, final URI url) {
+        if (exists(client, url)) {
+            deleteRecursive(client, url, new AtomicInteger(0));
+        }
+    }
+
+    static void deleteRecursive(final SolidSyncClient client, final URI url, final AtomicInteger depth) {
```
acoburn, line 226: I would avoid the use of a shared AtomicInteger here; instead, you can just have a `final int depth` and perform arithmetic operations on that in the recursive calls

edwardsph: Yes, the AtomicInteger was useful in an asynchronous context

edwardsph (commented):
Taking a step back and looking at how you manage test containers, this is what I see:

If `PUBLIC_RESOURCE_PATH` doesn't exist then you have
`  publicTestContainerURI = podUrl/test-UUID/`
and you rely on the ACR of the pod root being right.

If it does exist, then you have
`  publicTestContainerURI = podUrl/PUBLIC_RESOURCE_PATH/test-UUID/`
Then you create `podUrl/PUBLIC_RESOURCE_PATH/` and set up the ACR and either tear down this container at the end or try and ensure it is empty at the start. Since this is a fixed location, it will clash with other test runs.

Why not make that container unique too so you create `podUrl/PUBLIC_RESOURCE_PATH-UUID/`. That would isolate your tests but still use a defined prefix to make the test container identifiable (if that was the reason for allowing someone to set `PUBLIC_RESOURCE_PATH`. I may not have understood why you want someone to define the test container outside the tests.

The downside of my approach is that if something goes wrong with teardown, you can be left with old test containers but they will not interfere with anything and you can have a separate housekeeping process to clean them out. However, I think you will find that once the tests are more stable you will have less of a problem with this.

integration/base/src/main/java/com/inrupt/client/integration/base/CoreModulesResource.java
```diff
@@ -117,50 +113,42 @@ static void setup() {
...
                 .build();
 
-            publicContainerURI = URIBuilder.newBuilder(URI.create(podUrl))
-                    .path(PUBLIC_RESOURCE_PATH + "/").build();
-            createAuthenticatedClient();
-            createContainer(publicContainerURI);
-            prepareACR(publicContainerURI);
-        }
+        //if a tests fails it can be that the cleanup was not properly done, so we do it here too
+        Utils.deleteContentsRecursively(localAuthClient, publicContainerURI);
```
edwardsph, line 142: This line becomes redundant since you are now defining a unique new  `publicContainerURI`, so it doesn't yet exist and cannot have any contents. It is created on the next line. There may be other cases like this to check for.

integration/base/src/main/java/com/inrupt/client/integration/base/Utils.java
timea-solid: However, I cannot use it as `final int depth` but I do need a final in the lambda later. I did not find a better solution.

integration/base/src/main/java/com/inrupt/client/integration/base/AccessGrantScenarios.java
```diff
@@ -479,6 +449,10 @@ void accessGrantGetRdfTest(final Session resourceOwnerSession, final Session req
         final SolidSyncClient resourceOwnerClient = SolidSyncClient.getClientBuilder()
             .build().session(resourceOwnerSession);
 
+        final URI testRDFresourceURI = URIBuilder.newBuilder(privateContainerURI)
+                .path("resource-accessGrantGetRdfTest.ttl")
+                .build();
+
         try (final SolidRDFSource resource = new SolidRDFSource(testRDFresourceURI, null, null)) {
             assertDoesNotThrow(() -> resourceOwnerClient.create(resource));
```
NSeydoux, line 457: A pattern I typically use with the JS client for such case, where the resource URI is irrelevant to the client (but still needs to be tracked) is to get the server to create the resource by POSTing to the parent container, which URL the client does care about (here that would be `privateContainerURI`). Is this a pattern that would be possible with the JCL, or does the target URL needs to be known before creating a resource?

acoburn: In the high-level client, we exclusively use PUT for create and update. POST and PATCH are usable directly in the low-level client.

NSeydoux: It's absolutely not critical for GA, but would something like `createChild` make sense in the `SolidClient` ?

acoburn (approved):

PR #577, opened 2023-07-07
JCL-394: AccessRequest builder

acoburn:
This adds the implementation code for a builder pattern for access requests. In particular, this makes it possible to set an issuance date in the future, which is otherwise not possible in the current API.

The first commit just adds the boilerplate code for the new `AccessRequest.RequestParameter` class and connects that to the `AccessGrantClient.requestAccess` method (adding a new overload). The `getIssuedAt` value is then propagated through the rest of the methods (e.g. `grantAccess()` and `denyAccess()`).

Note: I re-ordered the parameters for some internal (private) methods to group the arguments into a more logical order. This does not affect the public API.

Some items still needed here include:

- [x] javadocs
- [x] tests

The way this would be used is like so:

```java

var req = AccessRequest.RequestParameters.newBuilder()
        .recipient(agent)
        .resource(uri)
        .modes(modes)
        .purposes(purposes)
        .expiration(expiration)
        .issuedAt(issuance).build();

agclient.requestAccess(req);
```

### DIFF
--- a/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessRequest.java
+++ b/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessRequest.java
@@ -30,6 +30,8 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
+import java.time.Instant;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


timea-solid (approved):

timea-solid:
Looks good to me. 
CI fails because of other known CI problems. (once we merge my PR, we should see clearer any problem).

acoburn:
@NSeydoux you are right, we're missing a static method: `newBuilder()`, which I would have seen as soon as I started writing tests, and which I will get to today :-)

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessRequest.java
```diff
@@ -133,4 +135,250 @@ static AccessRequest parse(final String serialization) throws IOException {
...
+            /**
+             * Set multiple purposes for the access request operation.
+             *
+             * <p>Note: A null value will clear all existing purpose values
+             *
+             * @param purposes the purpose identifiers, may be {@code null}
+             * @return this builder
+             */
+            public Builder purposes(final Collection<URI> purposes) {
+                if (purposes != null) {
```
NSeydoux, line 339: I don't think there is a test case for a collection cleared this way.

NSeydoux (approved):

PR #579, opened 2023-07-10
Re-enable integration tests using httpclient

NSeydoux:
These tests have some flakyness, but a re-run seems to be sufficient to address it, in addition to the ongoing work to fix test flakyness in general.


NSeydoux:
This currently introduces flakyness, we have a ticket to investigate later.

PR #588, opened 2023-07-10
JCL-371: Remove rdf-legacy dependency from tests

NSeydoux:
`rdf-legacy` only makes sense in a Java 8 context, it isn't necessary anywhere else. This depends on https://github.com/inrupt/solid-client-java/pull/575.


NSeydoux:
Merged into #575

PR #590, opened 2023-07-12
JCL-326: Remove deprecated methods and classes

acoburn:
This removes all deprecated classes and methods. The test code is also adjusted to accommodate these changes: in some cases I just remove the offending line and in other (where all deprecated methods were grouped together) I remove the entire test case, but generally, I tried to keep the existing test code intact.


timea-solid (approved):
It looks good to me. I am just wondering if it is enough to delete the deprecated from beta to beta version.

acoburn:
Technically, if we were to release a new Beta version, we'd want to reverse this commit, but since that's not the plan, I think we're good to go. If we do need to issue a new beta version, there are things we can do.

PR #591, opened 2023-07-12
JCL-422: Performance testing support

timea-solid:
Added a perf test in UMA on GET resource with grant on localhost.


acoburn:
Would it be possible to have the perf tests run in a separate GH Action step (i.e. in parallel) rather than after the e2e tests? That would shorten the wallclock time of the tests and make it easier to re-run, if necessary

timea-solid:
> Would it be possible to have the perf tests run in a separate GH Action step (i.e. in parallel) rather than after the e2e tests? That would shorten the wallclock time of the tests and make it easier to re-run, if necessary

Absolutely. And we don't nee dot run them on Java 11 also.

acoburn (approved):
LGTM

PR #593, opened 2023-07-13
Clean up sonar warnings

acoburn:
This cleans up some sonar/PMD warnings, including:

* Removing or using otherwise unused private fields
* Returning parameterized (rather than raw) types, where appropriate
* Use `@Override` where appropriate
* Avoid using the same variable name in a lambda that is used at the class level
* Replace the use of some deprecated methods in the Jena (test) module

### DIFF
--- a/openid/src/main/java/com/inrupt/client/openid/OpenIdSession.java
+++ b/openid/src/main/java/com/inrupt/client/openid/OpenIdSession.java
@@ -243,7 +243,7 @@ public final class OpenIdSession implements Session {
             final Request request, final Set<String> algorithms) {
         return auth.authenticate(this, request, algorithms)
                 .thenApply(cred-> {
-                    if (cred!= null) {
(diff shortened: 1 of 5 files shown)


timea-solid (approved):

openid/src/main/java/com/inrupt/client/openid/OpenIdSession.java
```diff
@@ -242,12 +242,12 @@ public void reset() {
     public CompletionStage<Optional<Credential>> authenticate(final Authenticator auth,
             final Request request, final Set<String> algorithms) {
         return auth.authenticate(this, request, algorithms)
-                .thenApply(credential -> {
-                    if (credential != null) {
+                .thenApply(cred-> {
```
timea-solid, line 245:
```suggestion
                .thenApply(cred -> {
```

timea-solid (approved):

PR #609, opened 2023-07-21
JCL-354: Remove maven site from project

acoburn:
This removes the maven site publication from this project, leaving only the publication of Javadocs


acoburn:
I have a better idea for this.

PR #611, opened 2023-07-21
JCL-98: Add maven enforcer plugin

acoburn:
This adds the maven enforcer plugin, which helps us ensure that the maven project structure follows best practices.


matthieubosquet (approved):
Seems like a great idea!

timea-solid (approved):

PR #622, opened 2023-07-25
JCL-264: use HTTPS on storage & added custom client which trusts all certs in e2e tests

timea-solid:
This PR is a preparation of e2e tests to work on https (the storage mock).
This requires also that the test use a custom client which trusts all certificates. This is important when running these tests on live servers.


timea-solid:
Updates this PR does:
* only starts the mock servers if there is no configuration pointing to a live servers webID;
* wires in a custom OkHttp client which accepts all certificates (needed on live servers) -> this was introduced as a submodule to the integration module. For httpClient we will also need a dedicated custom client later when we activate httpClient on the tests too.

acoburn (approved):
LGTM

PR #653, opened 2023-08-14
JCL-431: Improve containment validation

### LINKED ISSUE #634: getResources can include non-child resources
### Describe the bug

In PR #571 , a validation was added to exclude non-child resources, however the validation misses some edge cases.

### Expected behavior

`getResources()` from a container `/foo/` should not include `/foo/#bar` as a child (and anything else that is not a direct child resource).

### Actual behavior

It includes `/foo/#bar` as a child

### How to reproduce?

I've tested it with these lines in container.ttl and the unit tests:

```text/turtle
# These containment triples should not be included in a getResources response
<>
    ldp:contains <https://example.com/other> , <newContainer/child> , <> , <./>, <child/..>, <#hashtag>, <../> .
```

The test case failed because the hashtag was not filtered:

```
[ERROR]   SolidClientTest.testLowLevelSolidContainer » Completion org.opentest4j.AssertionFailedError: expected: <[http://localhost:36023/container/newContainer/, http://localhost:36023/container/test.txt, http:
//localhost:36023/container/test2.txt]> but was: <[http://localhost:36023/container/newContainer/, http://localhost:36023/container/test.txt, http://localhost:36023/container/test2.txt, http://localhost:36023/co
ntainer/#hashtag]>
```

### Output of `java -version`

_No response_

### Maven version

_No response_

### Cause

The problem likely is, that in the verifyContainmentIri method, the `normalizedPath.isEmpty()` is called as `"#hashtag".isEmpty()`. In other words, it thinks that the "#hashtag" is a child resource, rather than just a hashtag.

### Suggested fix

I'm not aware how to do this in Java, but I'd leave the path parsing to a library that already has all the edge cases built in (hashtag, query params, and not sure if there's more). Similar to how in solid-client-js they use the built-in `child.pathname` rather than manually parsing the path segments.

acoburn:
Resolves #634 

This improves containment validation by considering more cases to be handled, specifically those containment triples that include fragment and query components.

The test cases have been expanded to include the additional cases.

The implementation makes more extensive use of normalized Java URI objects to achieve the improved validation.

### DIFF
--- a/solid/src/main/java/com/inrupt/client/solid/SolidContainer.java
+++ b/solid/src/main/java/com/inrupt/client/solid/SolidContainer.java
@@ -82,9 +82,9 @@ public class SolidContainer extends SolidRDFSource {
      * @return the contained resources
      */
     public Set<SolidResource> getResources() {
-        final String container = normalize(getIdentifier());
         // As defined by the Solid Protocol, containers always end with a slash.
-        if (container.endsWith("/")) {
+        if (getIdentifier().getPath().endsWith("/")) {
+            final String container = normalize(getIdentifier());
             final Node node = new Node(rdf.createIRI(getIdentifier().toString()), getGraph());
             try (final Stream<Node.TypedNode> stream = node.getResources()) {
                 return stream.filter(child -> verifyContainmentIri(container, child)).map(child -> {
@@ -100,14 +100,14 @@ public class SolidContainer extends SolidRDFSource {
 
     @Override
     public ValidationResult validate() {
-        // Get the normalized container URI
-        final String container = normalize(getIdentifier());
         final List<String> messages = new ArrayList<>();
         // Verify that the container URI path ends with a slash
-        if (!container.endsWith("/")) {
+        if (!getIdentifier().getPath().endsWith("/")) {
             messages.add("Container URI does not end in a slash");
         }
 
+        // Get the normalized container URI
+        final String container = normalize(getIdentifier());
         // Verify that all ldp:contains triples align with Solid expectations
         getGraph().stream(null, rdf.createIRI(LDP.contains.toString()), null)
             .collect(Collectors.partitioningBy(verifyContainmentTriple(container)))
@@ -119,10 +119,6 @@ public class SolidContainer extends SolidRDFSource {
             return new ValidationResult(true);
         }
         return new ValidationResult(false, messages);
-    }
-
-    static String normalize(final IRI iri) {
-        return normalize(URI.create(iri.getIRIString()));
     }
 
     static String normalize(final URI uri) {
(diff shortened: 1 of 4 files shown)


solid/src/main/java/com/inrupt/client/solid/SolidContainer.java
```diff
@@ -82,9 +82,9 @@ public SolidContainer(final URI identifier, final Dataset dataset, final Metadat
      * @return the contained resources
      */
     public Set<SolidResource> getResources() {
-        final String container = normalize(getIdentifier());
         // As defined by the Solid Protocol, containers always end with a slash.
-        if (container.endsWith("/")) {
+        if (getIdentifier().getPath().endsWith("/")) {
+            final String container = normalize(getIdentifier());
```
langsamu, line 87: Newly, the condition is not normalized. This seems intentional.

But what happens with URIs of the form `x/.` and `x/y/..`?
Maybe we should normalize before checking whether it ends in a slash.

[Example](https://www.mycompiler.io/view/2W9PVVdf5DM)

langsamu: `getIdentifier` from `RDFSource` via `SolidRDFSource` is a `URI` and `getPath` is not normalized.

solid/src/main/java/com/inrupt/client/solid/SolidContainer.java
```diff
@@ -100,14 +100,14 @@ public Set<SolidResource> getResources() {
...
-        final String container = normalize(getIdentifier());
         final List<String> messages = new ArrayList<>();
         // Verify that the container URI path ends with a slash
-        if (!container.endsWith("/")) {
+        if (!getIdentifier().getPath().endsWith("/")) {
             messages.add("Container URI does not end in a slash");
         }
 
+        // Get the normalized container URI
+        final String container = normalize(getIdentifier());
```
langsamu, line 110: As above.

Maybe worth extracting something like `isContainer` for readability and consistency.

solid/src/main/java/com/inrupt/client/solid/SolidContainer.java
```diff
@@ -145,22 +141,43 @@ static Predicate<Triple> verifyContainmentTriple(final String container) {
     }
 
     static boolean verifyContainmentIri(final String container, final IRI object) {
-        if (!object.getIRIString().startsWith(container)) {
-            // Out-of-domain containment triple object
+
+        // 1 URI Structure Tests
```
langsamu, line 145: What do these numbers refer to? I looked in several place I imagined would be relevant, but haven't found anything like this.

solid/src/main/java/com/inrupt/client/solid/SolidContainer.java
acoburn: I've updated this to normalize the URI path

solid/src/main/java/com/inrupt/client/solid/SolidContainer.java
acoburn: done in https://github.com/inrupt/solid-client-java/pull/653/commits/2b4b91696bc4aa7f19774094d1f170183a16c9ac

solid/src/main/java/com/inrupt/client/solid/SolidContainer.java
acoburn: That was more for my own organization. I've backed that out

timea-solid (approved):

PR #668, opened 2023-08-31
JCL-434: Upgrade wiremock to 3.0

acoburn:
This upgrades wiremock to the 3.0 release series, updating the corresponding groupId and artifactId values.


timea-solid (approved):
It is curious how it does not show the <wiremock.version> change itself. But I see it is 3.0.0.
Nice cleanup.

PR #723, opened 2023-10-12
Maven logging dependency cleanup

acoburn:
This cleans up the maven configuration for logging. Specifically, there are cases where the slf4j-api is used in modules but the dependency is not listed explicitly. This can cause version mismatches when combined with wiremock, leading to log output not being produced.

Secondly, `wiremock-standalone` is being used in cases where it's not necessary (and we should generally prefer `wiremock` over `wiremock-standalone`).

Third, with the better wiremock output, there are some HTTP paths that are not being properly matched, but will be with this change.

### DIFF
--- a/access-grant/pom.xml
+++ b/access-grant/pom.xml
@@ -34,6 +34,11 @@
       <artifactId>commons-io</artifactId>
       <version>${commons.io.version}</version>
     </dependency>
+    <dependency>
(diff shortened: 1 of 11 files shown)


acoburn:
Dependency check failure addressed via #724

timea-solid (approved):

PR #740, opened 2023-10-16
Use more idiomatic dataset manipulation

acoburn:
The integrations tests manipulate RDF datasets in a somewhat non-idiomatic way. This reworks those tests to be more concise and (arguably) understandable.


NSeydoux (approved):
I don't have enough perpective to talk to the idiomacy of the changes, but I found the code easier to read indeed.

PR #774, opened 2023-11-14
JCL-435: Add headers data to core API

acoburn:
This add a `getHeaders()` method to the `com.inrupt.client.Resource` interface. The `SolidClient` then will add response headers to any Resource subtypes.

### DIFF
--- a/solid/src/main/java/com/inrupt/client/solid/SolidClient.java
+++ b/solid/src/main/java/com/inrupt/client/solid/SolidClient.java
@@ -138,14 +138,12 @@ public class SolidClient {
                 } else {
                     final String contentType = response.headers().firstValue(CONTENT_TYPE)
                         .orElse("application/octet-stream");
-                    final Metadata metadata = SolidResourceHandlers.buildMetadata(response.uri(),
-                            response.headers());
                     try {
                         // Check that this is an RDFSoure
                         if (RDFSource.class.isAssignableFrom(clazz)) {
                             final Dataset dataset = SolidResourceHandlers.buildDataset(contentType, response.body(),
                                     identifier.toString()).orElse(null);
-                            final T obj = construct(identifier, clazz, dataset, metadata);
+                            final T obj = construct(identifier, clazz, dataset, response.headers());
                             final ValidationResult res = RDFSource.class.cast(obj).validate();
                             if (!res.isValid()) {
                                 throw new DataMappingException(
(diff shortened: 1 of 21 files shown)


timea-solid (approved):
Interesting, I suppose this case out of the education materials.

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
```diff
@@ -388,30 +386,54 @@ <T extends Resource> Function<Response<byte[]>, CompletionStage<T>> handleRespon
...
+        // Fall back to an arity-2 ctor
+        return clazz.getConstructor(URI.class, Dataset.class)
+                    .newInstance(identifier, dataset);
     }
 
     static <T extends Resource> T construct(final URI identifier, final Class<T> clazz,
-            final String contentType, final InputStream entity, final Metadata metadata)
+            final String contentType, final InputStream entity, final Headers headers)
             throws ReflectiveOperationException {
+        // First try an arity-3 ctor with headers
```
langsamu, line 416:
```suggestion
        // First try an arity-4 ctor with headers
```

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
```diff
@@ -388,30 +386,54 @@ <T extends Resource> Function<Response<byte[]>, CompletionStage<T>> handleRespon
...
         try {
-            // First try an arity-4 ctor
+            return clazz.getConstructor(URI.class, String.class, InputStream.class, Headers.class)
+                .newInstance(identifier, contentType, entity, headers);
+        } catch (final NoSuchMethodException ex) {
+            // no-op
+        }
+
+        // Next try an arity-3 ctor with metadata
```
langsamu, line 424:
```suggestion
        // Next try an arity-4 ctor with metadata
```

solid/src/main/java/com/inrupt/client/solid/SolidResourceReference.java
```diff
@@ -49,6 +53,7 @@ public SolidResourceReference(final URI identifier, final Metadata metadata) {
             this.metadata = metadata;
         }
         this.entity = new ByteArrayInputStream(new byte[0]);
+        this.headers = Headers.of(new HashMap<>());
```
langsamu, line 56: `Headers.of(new HashMap<>())` is used so many times, maybe it's worth hiding behind a new `Headers.of()` (like `Set.of()` etc.) or `Headers.empty()`.

Thought `Set.of()` and the like tend to be immutable, so maybe a different name.

acoburn: Good idea. I added a `Headers.empty()` method in https://github.com/inrupt/solid-client-java/pull/774/commits/5927dbe302a42a60dfa7d2424e2ea69504d189a5

PR #832, opened 2023-12-01
Add support for dependency-check 9.x using an NVD API key

acoburn:
This makes use of an NVD API key and upgrades to version 9.x of the dependency-check plugin


acoburn:
The NVD API seems a little flaky still. Before merging this, I'd like to see evidence that the API is more stable

timea-solid (approved):

PR #838, opened 2023-12-04
JCL-446: config application metadata request tests

timea-solid:
Introduced 2 new configuration options:
* `inrupt.test.request-metadata.feature` can be `true` or `false`  (default is false) to signal if a env (prod or dev) should run the tests related to the feature.
* `inrupt.test.request-metadata-headers-that-propagate` is a list of strings such as `somecid, x-request-id` which are the headers active to propagate on the servers.

This PR activates the application request metadata tests only on dev2.2 - see CI config.
We only added some very basic tests, to be able to setup the config. More relevant tests will be added in a follow up PR:

### DIFF
--- a/integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
+++ b/integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
@@ -137,7 +137,7 @@ public class ApplicationRequestMetadataScenarios {
         LOGGER.info("Integration Test - Request and response headers match for a successful authenticated request");
 
         final Headers applicationHeaders = Headers.of(
-                Map.of("somecid", List.of("a6d87d0e-2454-4501-8110-ecc082aa975f"))
+                Map.of("someblabla", List.of("a6d87d0e-2454-4501-8110-ecc082aa975f"))
         );
 
         final SolidClient authClient = SolidClient.getClientBuilder()
--- a/integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
+++ b/integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
@@ -263,8 +263,8 @@ public class ApplicationRequestMetadataScenarios {
         // Create a new resource and check response headers
         final URI resourceUri = URI.create(resourceName);
         try ( var resource = authClient.create(new SolidRDFSource(resourceUri, dataset))) {
-            for (String value: REQUEST_METADATA_HEADERS_THAT_PROPAGATE) {
-                assertEquals(headers.get(value), resource.getHeaders().allValues(value));
+            for (String header: REQUEST_METADATA_HEADERS_THAT_PROPAGATE) {
+                assertEquals(headers.get(header), resource.getHeaders().allValues(header));
             }
             assertTrue(resource.getHeaders().allValues("someblabla").isEmpty());
         }
--- a/integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
+++ b/integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
@@ -21,8 +21,11 @@
 package com.inrupt.client.integration.base;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.inrupt.client.Headers;
+import com.inrupt.client.Request;
+import com.inrupt.client.Response;
 import com.inrupt.client.auth.Credential;
 import com.inrupt.client.auth.Session;
 import com.inrupt.client.openid.OpenIdException;
@@ -33,6 +36,7 @@ import com.inrupt.client.solid.SolidRDFSource;
 import com.inrupt.client.solid.SolidSyncClient;
 import com.inrupt.client.spi.RDFFactory;
 import com.inrupt.client.util.URIBuilder;
+import com.inrupt.client.vocabulary.LDP;
 import com.inrupt.client.webid.WebIdProfile;
 
 import java.net.URI;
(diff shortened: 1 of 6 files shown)


integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
```diff
@@ -0,0 +1,298 @@
...
+
+    private static final RDF rdf = RDFFactory.getInstance();
+
+    //feature is deactivated by default
+    private static final String REQUEST_METADATA_FEATURE = config
+            .getOptionalValue("inrupt.test.request-metadata.feature", String.class)
+            .orElse("false");
+
+    private static final String[] REQUEST_METADATA_HEADERS_THAT_PROPAGATE = config
+            .getValue("inrupt.test.request-metadata-headers-that-propagate", String[].class);
```
NSeydoux, line 76: I would have expected this to be optional. What is the value when the feature isn't being tested?

integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    private static boolean featureIsActive() {
+        return REQUEST_METADATA_FEATURE.equals("true");
+    }
+
+    private Map<String, List<String>> fillHeaders(final String headerValue) {
+        final var headers = new HashMap<String, List<String>>();
+        for (String value: REQUEST_METADATA_HEADERS_THAT_PROPAGATE) {
+            headers.put(value, List.of(headerValue + "-" + UUID.randomUUID()));
```
NSeydoux, line 294:
```suggestion
        for (String header: REQUEST_METADATA_HEADERS_THAT_PROPAGATE) {
            headers.put(header, List.of(headerValue + "-" + UUID.randomUUID()));
```

integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
```diff
@@ -0,0 +1,298 @@
...
+
+        for (String value: REQUEST_METADATA_HEADERS_THAT_PROPAGATE) {
+            assertEquals(headers.get(value), res.headers().allValues(value));
+        }
+    }
+
+    @ParameterizedTest
+    @MethodSource("provideSessions")
+    @DisplayName(" " +
+            "Request and response headers match for a successful authenticated request")
```
NSeydoux, line 198: The display name doesn't specify whether this is high or low-level client, I don't know if it should.

integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
```diff
@@ -0,0 +1,298 @@
...
+            }
+            assertTrue(response.getHeaders().allValues("someblabla").isEmpty());
+        });
+    }
+
+    @ParameterizedTest
+    @MethodSource("provideSessions")
+    @DisplayName(" " +
+            "Request and response headers match for a successful authenticated request")
+    void requestResponseMatchOnAuthRequestHighLevelSyncClientTest(final Session session) {
```
NSeydoux, line 238: The prefixes aren't in the same order as the previous test: should this be `requestResponseMatchOnAuthRequestSyncHighLevelClientTest`, or the other one be `requestResponseMatchOnAuthRequestHighLevelAsyncClientTest`?

integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
```diff
@@ -0,0 +1,298 @@
...
+
+        // Populate data for this resource
+        final Dataset dataset = rdf.createDataset();
+        dataset.add(null, subject, predicate, objectTrue);
+
+        // Create a new resource and check response headers
+        final URI resourceUri = URI.create(resourceName);
+        try ( var resource = authClient.create(new SolidRDFSource(resourceUri, dataset))) {
+            for (String value: REQUEST_METADATA_HEADERS_THAT_PROPAGATE) {
+                assertEquals(headers.get(value), resource.getHeaders().allValues(value));
```
NSeydoux, line 267:
```suggestion
            for (String header: REQUEST_METADATA_HEADERS_THAT_PROPAGATE) {
                assertEquals(headers.get(header), resource.getHeaders().allValues(header));
```

NSeydoux (approved):

integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
acoburn: Please make these config values optional. Otherwise, every developer who wants to build the code will need to add those configs to their environment. Plus, the build will fail in CD

integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
timea-solid: The display name needs to reference a scenario from RDF, when I add documentation. So it should be the same scenario regardless what client.
What I should improve is the logging to mention the client type.

integration/base/src/main/java/com/inrupt/client/integration/base/ApplicationRequestMetadataScenarios.java
```diff
@@ -0,0 +1,296 @@
...
+    private static final RDF rdf = RDFFactory.getInstance();
+
+    //feature is deactivated by default
+    private static final String REQUEST_METADATA_FEATURE = config
+            .getOptionalValue("inrupt.test.request-metadata.feature", String.class)
+            .orElse("false");
+
+    private static final String[] REQUEST_METADATA_HEADERS_THAT_PROPAGATE = config
+            .getOptionalValue("inrupt.test.request-metadata-headers-that-propagate", String[].class)
+            .orElse(new String[]{""});
```
acoburn, line 77: Better to initialize the array with either `new String[0]` or `new String[]{}`

PR #859, opened 2023-12-11
JCL-447: documented new configs

timea-solid:
Introduced 2 new configuration options:
* inrupt.test.request-metadata.feature can be true or false (default is false) to signal if a env (prod or dev) should run the tests related to the feature.
* inrupt.test.request-metadata-headers-that-propagate is a list of strings such as somecid, x-request-id which are the headers active to propagate on the servers.

### DIFF
--- a/integration/README.md
+++ b/integration/README.md
@@ -35,15 +35,10 @@ All the possible value are listed next:
 * `inrupt.test.webid`
 * `inrupt.test.access-grant.provider`
 * `inrupt.test.requester.webid`
-<<<<<<< Updated upstream
 * `inrupt.test.requester.client-id` // mandatory for running the AccessGrant scenarios
 * `inrupt.test.requester.client-secret` // mandatory for running the AccessGrant scenarios
-=======
-* `inrupt.test.requester.client-id` // mandatory
-* `inrupt.test.requester.client-secret` // mandatory
 * `inrupt.test.feature.request-metadata`
 * `inrupt.test.request-metadata-headers-that-propagate`
->>>>>>> Stashed changes
 
 Mandatory fields are:
 * `inrupt.test.client-id` & `inrupt.test.client-secret` are used to signal the server that this is a registered client acting as the owner of resources.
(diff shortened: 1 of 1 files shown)


integration/README.md
```diff
@@ -32,8 +35,15 @@ All the possible value are listed next:
 * `inrupt.test.webid`
 * `inrupt.test.access-grant.provider`
 * `inrupt.test.requester.webid`
+<<<<<<< Updated upstream
```
NSeydoux, line 38: These look like git conflict markers

integration/README.md
```diff
@@ -43,6 +53,9 @@ Optional fields are:
 * `inrupt.test.webid` is needed only if we want to run the integration tests on a live service. Otherwise, this property needs to be left out because it will be populated by the Mocked services with a mock username called `someuser`.
 * `inrupt.test.requester.webid` is only needed in the access grants test scenarios and can be also left empty because the Mocked services will create a username called `requester`.
 * `inrupt.test.auth-method` refers to the [client authentication](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication) method and has a default value of `client_secret_basic`. This value is used when this property is not provided.
+* `inrupt.test.feature.request-metadata` is a `true` or `false` config that signals if the feature is available on the server and only then runs the according scenarios. By default, this is set on `false`.
+* `inrupt.test.request-metadata-headers-that-propagate` if the above is `true`, the test scenarios need to know which headers the servers can propagate. This is a string comma separated list.
```
NSeydoux, line 57: Wouldn't it be more consistent to have `inrupt.test.feature.request-metadata` (as it currently is) and `inrupt.test.feature.request-metadata.headers-that-propagate`?

NSeydoux (changes requested):

integration/README.md
timea-solid: I was trying to make a difference between - flag for feature active or not and then the headers.

NSeydoux (approved):

PR #860, opened 2023-12-11
JCL-446: Configure integration tests

acoburn:
The current integration tests of this feature lead to failures in CD as well as failures in local (non-configured) development environments. This makes use of a more typical JUnit conditional, so that we don't end up with null pointer exceptions.


timea-solid (approved):
Yes I see my mistake now. Thanks for the fix.

PR #863, opened 2023-12-15
Build a shared cache for NVD API data

acoburn:
This builds a shared (daily) cache for use with the NVD API data. Once the cache is populated, it can be consumed by the dependency check plugin.


edwardsph (approved):
LGTM

PR #911, opened 2024-01-02
Add CVE-2023-51074 to suppression list

acoburn:
This suppresses a vulnerability in a transitive dependency of Wiremock that does not affect this project


edwardsph (approved):
LGTM

PR #964, opened 2024-01-29
JCL-451: Refactor access grant tests to reuse sessions

acoburn:
The access grant tests suffer from flakiness, related somehow to authentication. Rather than building OAuth2 client credential tokens for every test, this creates the Session objects once for the test suite and reuses it across test methods.


NSeydoux (approved):

acoburn:
Depends on #962

PR #1004, opened 2024-02-22
JCL-451: Serialize AccessGrant query requests

acoburn:
This appears to fix the flaky integration tests. It does it by changing the (questionable?) parallelization in the `AccessGrantClient::query` code. We previously had some issues with this in terms of the parallel processing of `InputStream` data.

In addition, as part of the investigation, I found that the HTTP client implementations do not properly respect `timeout` values set in a `Request` object. Setting a longer timeout did not, on its own, resolve this issue, but I kept that fix in this PR. If it would be preferred to pull that out into a separate PR, I can do that.

### DIFF
--- a/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
+++ b/access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
@@ -53,7 +53,6 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
-import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


langsamu (approved):
I understand what's going on. The elimination of parallelization makes sense.

But the async struggle after the change is not better than previously.
Blocking (joining the completable future) inside the iteration feels off.

Perhaps the way to go here is to build up a chain of client calls from the results of the query, then execure it.
I.e. reduce the collection of query results to a  (conceptually) linked list of futures, each representing one of the query result items, each cahained to the previous one, each producing an (ever growing) list of results.

This would take some time to author, but would remain properly asynchronous. And the calls would still end up being serial.

PR #1138, opened 2024-04-05
JCL-460: Initial support for RFC9457 Problem Details

NSeydoux:
This adds boilerplate for the Problem Details consumption. This PR intentionally doesn't update `SolidClient`, and rather gets the underlying infrastructure in place first, so that the more meaningful changes to the behavior are focused and not drowned in setup code.

Specifically, this creates the `ProblemDetails` class, and updates the exception class hierarchy so that the `SolidClientException` child classes take a `ProblemDetails` instance in their constructor. The `HttpStatus` enum will be used to provide a default title to the ProblemDetails derived from a non-RFC9457 compliant response.


api/src/main/java/com/inrupt/client/ClientHttpException.java
```diff
@@ -0,0 +1,53 @@
+package com.inrupt.client;
+
+import com.inrupt.client.spi.JsonService;
```
acoburn, line 3: The API class should not consume the SPI. It makes a circular dependency in the package graph

api/src/main/java/com/inrupt/client/ClientHttpException.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    private static ClientHttpException buildDefaultException(String message, int code, String status) {
+        return new ClientHttpException(
+                ProblemDetails.fromDefaultResponse(code, status),
+                message
+        );
+    }
+
+    public static ClientHttpException fromResponse(String message, int code, String status, Headers headers, InputStream body, JsonService jsonService) {
```
acoburn, line 42: This method should be removed

api/src/main/java/com/inrupt/client/ClientHttpException.java
NSeydoux: Fixed with 8f17f98e98 (and more generally with 85b25ebb8d399aabad4bfeca1ca00a28bad54675)

api/src/main/java/com/inrupt/client/ClientHttpException.java
NSeydoux: This has been moved to a private method of `SolidClient` with 72018e0c54cd8902d2709f775ae4766da65ed616.

solid/src/main/java/com/inrupt/client/solid/HttpStatus.java
```diff
@@ -0,0 +1,61 @@
...
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package com.inrupt.client.solid;
+
+import java.util.Arrays;
+import java.util.Optional;
+
+enum HttpStatus {
+    BAD_REQUEST(BadRequestException.STATUS_CODE, "Bad Request"),
```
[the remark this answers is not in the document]
NSeydoux: Agreed, that's what I wanted to do initially, but then the `SolidClientException::handle` `switch` statement complains that a union member variable (e.g. `HttpStatus.BAD_REQUEST.code`) is not a constant. I preferred to keep that switch statement rather than a large `if` for readability, but I don't have a strong opinion one way or another.

solid/src/main/java/com/inrupt/client/solid/HttpStatus.java
```diff
@@ -0,0 +1,61 @@
...
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package com.inrupt.client.solid;
+
+import java.util.Arrays;
+import java.util.Optional;
+
+enum HttpStatus {
```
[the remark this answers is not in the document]
langsamu: In my opinion, no.
Except when you want to communicate something important to ourselves.

solid/src/main/java/com/inrupt/client/solid/HttpStatus.java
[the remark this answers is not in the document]
NSeydoux: Done in d56823827d5a8a26d9aa6976fa44ab18c030ec27

NSeydoux:
Split into #1160 and #1161 to ease review, as this PR was growing too large.

PR #1157, opened 2024-04-12
JCL-402: Low-level client throws enriched exceptions

NSeydoux:
This is based on #1160, which should be reviewed first.

The `SolidClient` now uses the `throwOnError` body mapper to handle error responses, and provides a custom exception mapper to throw the appropriate specialized exception.

### DIFF
--- a/api/src/main/java/com/inrupt/client/Response.java
+++ b/api/src/main/java/com/inrupt/client/Response.java
@@ -97,6 +97,11 @@ public interface Response<T> {
         ByteBuffer body();
     }
 
+    /**
+     * Indicates whether a status code reflects a successful HTTP response.
+     * @param statusCode the HTTP response status code
+     * @return true if the status code is in the success range, namely [200, 299].
+     */
     static boolean isSuccess(final int statusCode) {
         return statusCode >= 200 && statusCode < 300;
     }
--- a/api/src/main/java/com/inrupt/client/Response.java
+++ b/api/src/main/java/com/inrupt/client/Response.java
@@ -163,49 +163,6 @@ public interface Response<T> {
             return responseInfo -> null;
         }
 
-        /**
-         * Throws on HTTP error using the provided mapper, or apply the provided body handler.
-         * @param handler the body handler to apply on non-error HTTP responses
-         * @param isSuccess a callback determining error cases
-         * @param exceptionMapper the exception mapper
-         * @return the body handler
-         * @param <T> the type of the body handler
-         */
-        public static <T> Response.BodyHandler<T> throwOnError(
-                final Response.BodyHandler<T> handler,
-                final Function<Response.ResponseInfo, Boolean> isSuccess,
-                final Function<Response.ResponseInfo, ClientHttpException> exceptionMapper
-        ) {
-            return responseinfo -> {
-                if (!isSuccess.apply(responseinfo)) {
-                    throw exceptionMapper.apply(responseinfo);
-                }
-                return handler.apply(responseinfo);
-            };
-        }
-
-        /**
-         * Throws on HTTP error, or apply the provided body handler.
-         * @param handler the body handler to apply on non-error HTTP responses
-         * @param isSuccess a callback determining error cases
-         * @return the body handler
-         * @param <T> the type of the body handler
-         */
-        public static <T> Response.BodyHandler<T> throwOnError(
-                final Response.BodyHandler<T> handler,
-                final Function<Response.ResponseInfo, Boolean> isSuccess
-        ) {
-            final Function<Response.ResponseInfo, ClientHttpException> defaultMapper = responseInfo ->
-                new ClientHttpException(
-                    "An HTTP error has been returned, with status code " + responseInfo.statusCode(),
-                    responseInfo.uri(),
-                    responseInfo.statusCode(),
-                    responseInfo.headers(),
-                    new String(responseInfo.body().array(), StandardCharsets.UTF_8)
-                );
-            return throwOnError(handler, isSuccess, defaultMapper);
-        }
-
         private BodyHandlers() {
             // Prevent instantiation
         }
--- a/api/src/main/java/com/inrupt/client/Response.java
+++ b/api/src/main/java/com/inrupt/client/Response.java
@@ -26,8 +26,6 @@ import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.net.URI;
 import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
-import java.util.function.Function;
 
 /**
  * An HTTP Response.
--- a/api/src/main/java/com/inrupt/client/Response.java
+++ b/api/src/main/java/com/inrupt/client/Response.java
@@ -97,6 +97,10 @@ public interface Response<T> {
         ByteBuffer body();
     }
 
+    static boolean isSuccess(final int statusCode) {
+        return statusCode >= 200 && statusCode < 300;
+    }
+
     /**
      * An interface for mapping an HTTP response into a specific Java type.
      * @param <T> the body type
@@ -196,8 +200,6 @@ public interface Response<T> {
                 );
             return throwOnError(handler, isSuccess, defaultMapper);
         }
-
-
 
         private BodyHandlers() {
             // Prevent instantiation
--- a/solid/src/main/java/com/inrupt/client/solid/SolidClient.java
+++ b/solid/src/main/java/com/inrupt/client/solid/SolidClient.java
@@ -22,14 +22,11 @@ package com.inrupt.client.solid;
 
 import com.inrupt.client.*;
 import com.inrupt.client.auth.Session;
-import com.inrupt.client.spi.JsonService;
-import com.inrupt.client.spi.ServiceProvider;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
-import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -127,7 +124,10 @@ public class SolidClient {
         final Request request = builder.build();
         return client.send(
                 request,
-                Response.BodyHandlers.throwOnError(Response.BodyHandlers.ofByteArray(), (r) -> isSuccess(r.statusCode()))
+                Response.BodyHandlers.throwOnError(
+                    Response.BodyHandlers.ofByteArray(),
+                    (r) -> isSuccess(r.statusCode())
+                )
             ).thenApply(response -> {
                 final String contentType = response.headers().firstValue(CONTENT_TYPE)
                     .orElse("application/octet-stream");
@@ -157,7 +157,7 @@ public class SolidClient {
                 if (exception instanceof ClientHttpException) {
                     throw SolidClientException.handle((ClientHttpException) exception);
                 }
-                throw new RuntimeException("Something went wrong reading "+request.uri(), exception);
+                throw new RuntimeException("Something went wrong reading " + request.uri(), exception);
             });
     }
 
@@ -277,18 +277,18 @@ public class SolidClient {
         headers.firstValue(USER_AGENT).ifPresent(agent -> builder.setHeader(USER_AGENT, agent));
 
         return client.send(
-                builder.build(),
-                Response.BodyHandlers.throwOnError(
-                    Response.BodyHandlers.ofByteArray(),
-                    (r) -> isSuccess(r.statusCode())
-                )
-            ).exceptionally(exception -> {
+            builder.build(),
+            Response.BodyHandlers.throwOnError(
+                Response.BodyHandlers.ofByteArray(),
+                (r) -> isSuccess(r.statusCode())
+            )
+        ).exceptionally(exception -> {
             if (exception instanceof ClientHttpException) {
                 throw SolidClientException.handle((ClientHttpException) exception);
             }
-            throw new RuntimeException("Something went wrong reading "+resource.getIdentifier(), exception);
+            throw new RuntimeException("Something went wrong reading " + resource.getIdentifier(), exception);
             // FIXME I don't understand why the following is required.
-        }).thenAccept(o -> {});
+        }).thenAccept(o -> { /* no-op */ });
     }
 
     /**
--- a/solid/src/main/java/com/inrupt/client/solid/SolidClient.java
+++ b/solid/src/main/java/com/inrupt/client/solid/SolidClient.java
@@ -382,7 +382,7 @@ public class SolidClient {
                     resource.getIdentifier(),
                     res.statusCode(),
                     res.headers(),
-                    new String(res.body())
+                    new String(res.body(), StandardCharsets.UTF_8)
                 );
             }
 
--- a/solid/src/main/java/com/inrupt/client/solid/SolidClient.java
+++ b/solid/src/main/java/com/inrupt/client/solid/SolidClient.java
@@ -140,7 +140,7 @@ public class SolidClient {
                 Response.BodyHandlers.throwOnError(
                     Response.BodyHandlers.ofByteArray(),
                     (r) -> Response.isSuccess(r.statusCode()),
-                    httpExceptionMapper("Reading resource " + request.uri() + " failed.")
+                    httpExceptionMapper("Reading resource failed.")
                 )
             ).thenApply(response -> {
                 final String contentType = response.headers().firstValue(CONTENT_TYPE)
--- a/solid/src/main/java/com/inrupt/client/solid/SolidClient.java
+++ b/solid/src/main/java/com/inrupt/client/solid/SolidClient.java
@@ -139,7 +139,7 @@ public class SolidClient {
                 request,
                 Response.BodyHandlers.throwOnError(
                     Response.BodyHandlers.ofByteArray(),
-                    (r) -> isSuccess(r.statusCode()),
+                    (r) -> Response.isSuccess(r.statusCode()),
                     httpExceptionMapper("Reading resource " + request.uri() + " failed.")
                 )
             ).thenApply(response -> {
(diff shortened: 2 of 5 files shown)


solid/src/main/java/com/inrupt/client/solid/SolidClient.java
```diff
@@ -64,11 +54,21 @@ public class SolidClient {
...
     private final boolean fetchAfterWrite;
+    private final JsonService jsonService;
 
     SolidClient(final Client client, final Headers headers, final boolean fetchAfterWrite) {
         this.client = Objects.requireNonNull(client, "Client may not be null!");
         this.defaultHeaders = Objects.requireNonNull(headers, "Headers may not be null!");
         this.fetchAfterWrite = fetchAfterWrite;
+        JsonService js;
+        try {
+            // It is acceptable for a SolidClient instance to be in a classpath without any immplementation for
```
NSeydoux, line 65:
```suggestion
            // It is acceptable for a SolidClient instance to be in a classpath without any implementation for
```

solid/src/main/java/com/inrupt/client/solid/SolidClientException.java
```diff
@@ -107,6 +116,62 @@ public String getBody() {
...
+            js = ServiceProvider.getJsonService();
+        } catch (IllegalStateException e) {
+            js = null;
+        }
+        SolidClientException.jsonService = js;
+        SolidClientException.jsonServiceInitialized = true;
+        return SolidClientException.jsonService;
+    }
+
+    public static SolidClientException fromErrorResponse(
```
NSeydoux, line 140: Arguably, moving this method here makes it possible to make the `handle` method private.

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
```diff
@@ -64,11 +57,22 @@ public class SolidClient {
...
+    private final JsonService jsonService;
 
     SolidClient(final Client client, final Headers headers, final boolean fetchAfterWrite) {
         this.client = Objects.requireNonNull(client, "Client may not be null!");
         this.defaultHeaders = Objects.requireNonNull(headers, "Headers may not be null!");
         this.fetchAfterWrite = fetchAfterWrite;
+
+        // It is acceptable for a SolidClient instance to be in a classpath without any implementation for
+        // JsonService, in which case the ProblemDetails exceptions will fallback to default and not be parsed.
+        JsonService js;
```
langsamu, line 69: Local is redundant.

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
```diff
@@ -134,8 +138,19 @@ public <T extends Resource> CompletionStage<T> read(final URI identifier, final
...
+                        response.body(),
+                        this.jsonService
+                    );
+                    throw SolidClientException.handle(
+                        "Unable to read resource at " + request.uri(),
+                        pd,
+                        response.uri(),
+                        response.headers(),
+                        new String(response.body())
+                    );
```
langsamu, line 153: There are three of these blocks here. They are very similar.

Could they be extracted?

solid/src/test/java/com/inrupt/client/solid/SolidClientTest.java
```diff
@@ -413,23 +415,366 @@ public int statusCode() {
...
+                Arguments.of(NotFoundException.class, 404, "Not Found"),
+                Arguments.of(MethodNotAllowedException.class, 405, "Method Not Allowed"),
+                Arguments.of(NotAcceptableException.class, 406, "Not Acceptable"),
+                Arguments.of(ConflictException.class, 409, "Conflict"),
+                Arguments.of(GoneException.class, 410, "Gone"),
+                Arguments.of(PreconditionFailedException.class, 412, "Precondition Failed"),
+                Arguments.of(UnsupportedMediaTypeException.class, 415, "Unsupported Media Type"),
+                Arguments.of(TooManyRequestsException.class, 429, "Too Many Requests"),
+                Arguments.of(InternalServerErrorException.class, 500, "Internal Server Error"),
+                Arguments.of(SolidClientException.class, 418, "Bad Request")
```
langsamu, line 442: Dislike mentioned earlier.

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
```diff
@@ -284,8 +299,19 @@ public <T extends Resource> CompletionStage<Void> delete(final T resource, final
             if (isSuccess(res.statusCode())) {
                 return null;
             } else {
-                throw SolidClientException.handle("Unable to delete resource", resource.getIdentifier(),
-                        res.statusCode(), res.headers(), new String(res.body(), UTF_8));
```
jholleran, line 288: Note the UTF_8 change here. Is this ok?

acoburn: Definitely want UTF_8. Otherwise it defaults to the system defined encoding

NSeydoux: Good catch, fixed with 43f3a02196

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
```diff
@@ -131,36 +135,37 @@ public <T extends Resource> CompletionStage<T> read(final URI identifier, final
         headers.firstValue(USER_AGENT).ifPresent(agent -> builder.setHeader(USER_AGENT, agent));
 
         final Request request = builder.build();
-        return client.send(request, Response.BodyHandlers.ofByteArray())
```
NSeydoux, line 134: This looks like a large change, but essentially the error branch of the initial `if` has been moved to the wrapping `throwOnError`

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
```diff
@@ -280,14 +285,15 @@ public <T extends Resource> CompletionStage<Void> delete(final T resource, final
...
-        });
+        return client.send(
+            builder.build(),
+            Response.BodyHandlers.throwOnError(
+                Response.BodyHandlers.ofByteArray(),
+                (r) -> isSuccess(r.statusCode()),
+                httpExceptionMapper("Deleting resource " + resource.getIdentifier() + " failed.")
+            )
+        )// FIXME I don't understand why the following is required.
+        .thenAccept(o -> { /* no-op */ });
```
NSeydoux, line 296: I don't know that this is the correct approach

solid/src/test/java/com/inrupt/client/solid/SolidClientTest.java
NSeydoux: Now removed with https://github.com/inrupt/solid-client-java/pull/1157/commits/c120c8b05e4d42cbc42cf35f82f9d57998aa4f50

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
NSeydoux: This is now handled by the exception mapper.

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
NSeydoux: This has now been removed. I had a compilation error because of this, which in retrospective was due to the member variable being immutable and set both in the `try` and `catch` blocks, which I thought should be okay as I wouldn't expect both assignments to be able to happen consecutively.

NSeydoux:
You'll notice that unlike what we discussed yesterday, I didn't end up using the `exceptionally` clause of the `CompletionStage` chain, and instead changed the `throwOnError` `BodyHandler` so that it accepts a custom exception mapper. This prevents from having to add a mapping from generic `ClientHttpException` to module-specific `SolidClientException`, which didn't provide much value.

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
langsamu: It's because the method signature is (async) `void`.

Previously (when error handling was here) the response from the body handler was mapped (`thenApply`) to nothing.

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
```diff
@@ -371,8 +377,13 @@ <T extends Resource> Function<Response<byte[]>, CompletionStage<T>> handleRespon
...
         return res -> {
             if (!isSuccess(res.statusCode())) {
-                throw SolidClientException.handle(message, resource.getIdentifier(),
-                        res.statusCode(), res.headers(), new String(res.body(), UTF_8));
+                throw SolidClientException.handle(
+                    message,
+                    resource.getIdentifier(),
+                    res.statusCode(),
+                    res.headers(),
+                    new String(res.body())
```
langsamu, line 385: UTF again?

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
```diff
@@ -280,14 +285,15 @@ public <T extends Resource> CompletionStage<Void> delete(final T resource, final
...
-                throw SolidClientException.handle("Unable to delete resource", resource.getIdentifier(),
-                        res.statusCode(), res.headers(), new String(res.body(), UTF_8));
-            }
-        });
+        return client.send(
+            builder.build(),
+            Response.BodyHandlers.throwOnError(
+                Response.BodyHandlers.ofByteArray(),
+                (r) -> isSuccess(r.statusCode()),
+                httpExceptionMapper("Deleting resource " + resource.getIdentifier() + " failed.")
```
jholleran, line 293: Potential PII in message. Should the message be left at the previous `Unable to delete resource`. The resource URI will be `SolidClientException`

NSeydoux: Removed with https://github.com/inrupt/solid-client-java/pull/1157/commits/c03c1d93fd1dfec6644150ecf299f5c027472e59

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
```diff
@@ -131,36 +135,37 @@ public <T extends Resource> CompletionStage<T> read(final URI identifier, final
...
-                        // Otherwise, create a non-RDF-bearing resource
-                        } else {
-                            return construct(request.uri(), clazz, contentType,
-                                    new ByteArrayInputStream(response.body()), response.headers());
+        return client.send(
+                request,
+                Response.BodyHandlers.throwOnError(
+                    Response.BodyHandlers.ofByteArray(),
+                    (r) -> Response.isSuccess(r.statusCode()),
+                    httpExceptionMapper("Reading resource " + request.uri() + " failed.")
```
jholleran, line 143: Do you want to remove the resource URI from this now?

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
```diff
@@ -280,14 +285,15 @@ public <T extends Resource> CompletionStage<Void> delete(final T resource, final
...
-        });
+        return client.send(
+            builder.build(),
+            Response.BodyHandlers.throwOnError(
+                Response.BodyHandlers.ofByteArray(),
+                (r) -> Response.isSuccess(r.statusCode()),
+                httpExceptionMapper("Unable to delete resource.")
+            )
+        )
+        .thenAccept(o -> { /* no-op */ });
```
jholleran, line 296: What does this thenAccept no-op do? Is it needed?

solid/src/test/java/com/inrupt/client/solid/SolidClientTest.java
```diff
@@ -413,23 +672,58 @@ public int statusCode() {
                         })
         );
         assertEquals(statusCode, exception.getStatusCode());
+        // On malformed response, the ProblemDetails should fall back to defaults.
```
jholleran, line 675: Does this mean the client will ignore weird or corrupted responses from the server?

solid/src/main/java/com/inrupt/client/solid/SolidClient.java
NSeydoux: It is needed to honor the function signature. I had a similar doubt/question :) : https://github.com/inrupt/solid-client-java/pull/1157#discussion_r1568766232

solid/src/test/java/com/inrupt/client/solid/SolidClientTest.java
NSeydoux: That is correct: any response that is not valid Problem Details, either because RFC9457 is not supported by the server at all or because the server returns a malformed response while having the `application/problem+json` `Content-Type` will have the same behavior of creating a default `ProblemDetails` object (only having the status code) that will be tacked on the exception.

solid/src/test/java/com/inrupt/client/solid/SolidClientTest.java
```diff
@@ -379,28 +380,286 @@ private static Stream<Arguments> testExceptionalResources() {
...
+        assertEquals(problemDetails.getType(), exception.getProblemDetails().getType());
+        assertEquals(problemDetails.getTitle(), exception.getProblemDetails().getTitle());
+        assertEquals(problemDetails.getStatus(), exception.getProblemDetails().getStatus());
+        assertEquals(problemDetails.getDetails(), exception.getProblemDetails().getDetails());
+        assertEquals(problemDetails.getInstance(), exception.getProblemDetails().getInstance());
+    }
+
+    private static Stream<Arguments> testRfc9457Exceptions() {
+        return Stream.of(
+                Arguments.of(
```
langsamu, line 501: Minor: The `arguments` static method on `Arguments` is the same as its `of` and reads a bit better.

solid/src/test/java/com/inrupt/client/solid/SolidClientTest.java
```diff
@@ -379,28 +380,286 @@ private static Stream<Arguments> testExceptionalResources() {
...
+                        BadRequestException.class,
+                        new ProblemDetails(
+                                URI.create("https://example.org/type"),
+                                "Bad Request",
+                                "Some details",
+                                400,
+                                URI.create("https://example.org/instance")
+                        )
+                ), Arguments.of(
```
langsamu, line 510: Minor: It would look better to my eyes if these were on a new line.

solid/src/test/java/com/inrupt/client/solid/SolidClientTest.java
```diff
@@ -379,28 +380,286 @@ private static Stream<Arguments> testExceptionalResources() {
...
+        assertEquals(problemDetails.getDetails(), exception.getProblemDetails().getDetails());
+        assertEquals(problemDetails.getInstance(), exception.getProblemDetails().getInstance());
+    }
+
+    private static Stream<Arguments> testRfc9457Exceptions() {
+        return Stream.of(
+                Arguments.of(
+                        BadRequestException.class,
+                        new ProblemDetails(
+                                URI.create("https://example.org/type"),
```
langsamu, line 504: I would extract what I can into locals of `testRfc9457Exceptions`.
Anything to make the test case rows more readable.

langsamu (approved):

NSeydoux:
Javadoc build currently failing because of https://github.com/maxcellent/javadoc.io/issues/188

solid/src/test/java/com/inrupt/client/solid/SolidClientTest.java
NSeydoux: Done in 0131730c0a76e351ecd22b312ee814969c80dd5c

solid/src/test/java/com/inrupt/client/solid/SolidClientTest.java
NSeydoux: Done in 8eed2118b465f22d61cb91b6c9c0e0b1be1247a3

solid/src/test/java/com/inrupt/client/solid/SolidClientTest.java
NSeydoux: Attempted improvement in 8eed2118b465f22d61cb91b6c9c0e0b1be1247a3

jholleran (approved):

NSeydoux:
I'm currently investigating the e2e test failures

solid/src/test/java/com/inrupt/client/solid/SolidClientTest.java
langsamu: Looks good.

api/src/main/java/com/inrupt/client/Response.java
```diff
@@ -97,6 +95,15 @@ interface ResponseInfo {
         ByteBuffer body();
     }
 
+    /**
+     * Indicates whether a status code reflects a successful HTTP response.
+     * @param statusCode the HTTP response status code
+     * @return true if the status code is in the success range, namely [200, 299].
+     */
+    static boolean isSuccess(final int statusCode) {
```
acoburn, line 103: Do we really need this method?

solid/src/test/java/com/inrupt/client/solid/SolidClientTest.java
```diff
@@ -366,41 +368,232 @@ <T extends SolidClientException> void testExceptionalResources(
...
+        final SolidContainer resource = new SolidContainer(URI.create("http://example.com"));
+
         final SolidClientException exception = assertThrows(
                 clazz,
                 () -> solidClient.handleResponse(resource, headers, "message")
                         .apply(new Response<byte[]>() {
                             @Override
                             public byte[] body() {
-                                return new byte[0];
+                                final ByteArrayOutputStream bos = new ByteArrayOutputStream();
```
acoburn, line 458: Use try-catch here

api/src/main/java/com/inrupt/client/Response.java
NSeydoux: It avoids some duplication, and it is a quite common pattern to have this affordance on HTTP response object: JS `Response` has a `.ok` getter, `OkHttp` has [`isSuccessful`](https://square.github.io/okhttp/5.x/okhttp/okhttp3/-response/is-successful.html)... I'm not strongly opposed to removing this, but I find it reads slightly better than checking the status range each time.

solid/src/test/java/com/inrupt/client/solid/SolidClientTest.java
NSeydoux: Fixed in 9e0967da68

PR #1159, opened 2024-04-14
JCL-402: HTTP error handling in JenaBodyHandlers

NSeydoux:
This is based on #1160, which should be reviewed first.

This deprecates the current methods in `JenaBodyHandlers`, and replaces them with similar method throwing an appropriate exception with error details on HTTP error instead of returning an empty dataset. This makes use of the composable throwing body handler now exposed by the `Response` interface of the `api` module. The new method now have `Jena` in their name for this not to be a breaking change: `ofModel` becomes `ofJenaModel`, etc.

### DIFF
--- a/jena/src/main/java/com/inrupt/client/jena/JenaBodyHandlers.java
+++ b/jena/src/main/java/com/inrupt/client/jena/JenaBodyHandlers.java
@@ -20,11 +20,13 @@
  */
 package com.inrupt.client.jena;
 
+import com.inrupt.client.ClientHttpException;
 import com.inrupt.client.Response;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
 
 import org.apache.jena.atlas.web.ContentType;
 import org.apache.jena.graph.Graph;
@@ -43,6 +45,18 @@ import org.apache.jena.riot.RDFLanguages;
 public final class JenaBodyHandlers {
 
     private static final String CONTENT_TYPE = "Content-Type";
+
+    private static void throwOnError(final Response.ResponseInfo responseInfo) {
+        if (!Response.isSuccess(responseInfo.statusCode())) {
+            throw new ClientHttpException(
+                    "Could not map to a Jena entity.",
+                    responseInfo.uri(),
+                    responseInfo.statusCode(),
+                    responseInfo.headers(),
+                    new String(responseInfo.body().array(), StandardCharsets.UTF_8)
+            );
+        }
+    }
 
     private static Model responseToModel(final Response.ResponseInfo responseInfo) {
         return responseInfo.headers().firstValue(CONTENT_TYPE)
@@ -75,10 +89,10 @@ public final class JenaBodyHandlers {
      * @return an HTTP body handler
      */
     public static Response.BodyHandler<Model> ofJenaModel() {
-        return Response.BodyHandlers.throwOnError(
-                JenaBodyHandlers::responseToModel,
-                (r) -> Response.isSuccess(r.statusCode())
-        );
+        return responseInfo -> {
+            JenaBodyHandlers.throwOnError(responseInfo);
+            return JenaBodyHandlers.responseToModel(responseInfo);
+        };
     }
 
     private static Graph responseToGraph(final Response.ResponseInfo responseInfo) {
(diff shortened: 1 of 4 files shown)


jena/src/main/java/com/inrupt/client/jena/JenaBodyHandlers.java
```diff
@@ -43,14 +47,28 @@
 public final class JenaBodyHandlers {
 
     private static final String CONTENT_TYPE = "Content-Type";
+    private static JsonService jsonService;
```
langsamu, line 50: Something feels wrong about the Jena module developing a dependency (even if weak) on the JSON module.

I follow why this is needed, but it feels off regardless.

jena/src/main/java/com/inrupt/client/jena/JenaBodyHandlers.java
```diff
@@ -65,12 +83,39 @@ public static Response.BodyHandler<Model> ofModel() {
...
+    public static Response.BodyHandler<Model> ofModel() {
+        return JenaBodyHandlers::responseToModel;
+    }
+
+    /**
+     * Populate a Jena {@link Model} with an HTTP response body.
+     *
+     * @return an HTTP body handler
+     */
+    public static Response.BodyHandler<Model> ofJenaModel() {
```
langsamu, line 100: Looking at this I think I have an idea of what's wrong with the new module dependency criticized above.

I'll try to argue that the body handler is not the right place to translate a response into problem details.
Whoever is invoking the body handler (our code, not users') should know whether the response is 'ok' (and thus eligible for handling) or 'not ok' (thus eligible for problem detail parsing instead).

I haven't exercised this theory in the context of the codebase, but it seems sound.

jena/src/main/java/com/inrupt/client/jena/JenaBodyHandlers.java
NSeydoux: With the change proposed in 2ed0a03ba49d8c5fac59e0f01ae37ee3e545798c, this dependency can be removed. I thought that approach would create weird dependency cycles, but it appears to be working fine.

jena/src/main/java/com/inrupt/client/jena/JenaBodyHandlers.java
NSeydoux: As we discussed yesterday, the body handler is provided to the HTTP client, at which layer there is no business logic about what is an error vs what isn't. I think this means there is no other place to handle error responses but the body handler itself.

jena/src/main/java/com/inrupt/client/jena/JenaBodyHandlers.java
```diff
@@ -44,13 +44,12 @@ public final class JenaBodyHandlers {
...
-    /**
-     * Populate a Jena {@link Model} with an HTTP response body.
-     *
-     * @return an HTTP body handler
-     */
-    public static Response.BodyHandler<Model> ofModel() {
-        return responseInfo -> responseInfo.headers().firstValue(CONTENT_TYPE)
+    private static Boolean isSuccess(final Response.ResponseInfo responseInfo) {
+        return responseInfo.statusCode() < 300;
+    }
```
langsamu, line 49: There are now three of these:
- Here,
- in int RDF4J counterpart and
- [in the SolidClient](https://github.com/inrupt/solid-client-java/blob/e3e9ba52a8902d9d09e5b7af55dd06dffe3690d7/solid/src/main/java/com/inrupt/client/solid/SolidClient.java#L448) (this one is slightly different).

I would love to see only one if you find a way.

jena/src/main/java/com/inrupt/client/jena/JenaBodyHandlers.java
```diff
@@ -65,12 +64,29 @@ public static Response.BodyHandler<Model> ofModel() {
     }
 
     /**
-     * Populate a Jena {@link Graph} with an HTTP response.
+     * Populate a Jena {@link Model} with an HTTP response body.
      *
      * @return an HTTP body handler
+     * @deprecated Use ofJenaModel instead for consistent HTTP error handling.
```
langsamu, line 70: Can we `@{link}` in `@deprecated`?

jena/src/main/java/com/inrupt/client/jena/JenaBodyHandlers.java
NSeydoux: https://github.com/inrupt/solid-client-java/pull/1157/commits/5f41ca04f15b8620436334a20c66b650fdf7db2e lifts the logic into the API module, and https://github.com/inrupt/solid-client-java/pull/1159/commits/f92ed37b357693e0b02e6dbda9d5b46d8b4caebf uses the shared function here.

langsamu (approved):

PR #1160, opened 2024-04-14
JCL-402: Data classes for HTTP Problem Details support

NSeydoux:
This creates the `ProblemDetails` class, which represents the content of an RFC9457-compliant HTTP error response. ~To help setting defaults for `ProblemDetails`, this also adds `HttpStatus` to the public API of the `api` module. Only the HTTP status are part of the API (they will be reused by error-specific exceptions), and the `HttpStatus` class also defines a package-private way to build a default Problem Details title based on the HTTP status code.~

`ProblemDetails` encapsulates the SPI lookup required to discover a JSON service used to deserialize RFC9457 error responses. If no JSON parser is available, `ProblemDetails` falls back to providing a minimal object with all its fields nulled except for the type set to a default value and the error code set to the value associated with the response. Unlike what was suggested in the initial design, no default is applied to the title based on the HTTP status phrase, because not all the HTTP client libraries we support expose sur HTTP status phrase.

The tests for `ProblemDetails` parsing are in the solid-client module, because it will need the additional json service anyways in the test dependencies for testing the rest of this feature, and putting these tests in the api module creates a circular dependency.

The `Response` interface from the `api` module now also exposes a new `BodyHandler` that will throw on HTTP error. This body handler is composable with another body handler that handles success cases, and projects the HTTP response to the desirable object. Both the exception mapper and the success criterion are configurable. This new body handler throws a `ClientHttpException`, which is a new Exception type parent of `SolidClientException`. The HTTP-specific members of `SolidClientException` have been lifted up to `ClientHttpException`.


api/src/main/java/com/inrupt/client/HttpStatus.java
```diff
@@ -0,0 +1,80 @@
...
+                .map(knownStatus -> knownStatus.message)
+                .orElseGet(() -> {
+                    // If the status is unknown, default to 400 for client errors and 500 for server errors
+                    if (statusCode >= 400 && statusCode <= 499) {
+                        return BAD_REQUEST.message;
+                    }
+                    return INTERNAL_SERVER_ERROR.message;
+                });
+        }
+    }
```
langsamu, line 75: I don't like this logic.
While I think I understand why it would seem helpful, I disagree with it.

Other implementations of HTTP status classes that I know of choose either to return nulls or to throw when they encounter unknown statuses.

The default chosen here is very opinionated and I don't think that a library like this one, a fairly low-level HTTP client library, should make such decisions about responses from the server.

I might even argue the it is always the server who is responsible for the content of HTTP status codes and phrases, so our library should just honor the ones generated by the server.

[Another example of prior art](https://github.com/inrupt/enterprise-authorization/pull/801#discussion_r1562485606) is the JBoss RESTEasy implementation of JAX-RS (I think), that has a special HTTP status phrase 'Unknown Code' for some cases.

api/src/main/java/com/inrupt/client/ProblemDetails.java
```diff
@@ -0,0 +1,132 @@
...
+            final String title = Optional.ofNullable((String) pdData.get("title"))
+                    .orElse(HttpStatus.StatusMessages.getStatusMessage(statusCode));
+            final String details = (String) pdData.get("details");
+            final URI type = Optional.ofNullable((String) pdData.get("type"))
+                    .map(URI::create)
+                    .orElse(null);
+            final URI instance = Optional.ofNullable((String) pdData.get("instance"))
+                    .map(URI::create)
+                    .orElse(null);
+            // Note that the status code is disregarded from the body, and reused from the HTTP response directly,
```
langsamu, line 119: We could say that the mentioned requirement is for the server.
If the server decides to err, I think we should return what the server said.

api/src/main/java/com/inrupt/client/ProblemDetails.java
```diff
@@ -0,0 +1,132 @@
...
+                HttpStatus.StatusMessages.getStatusMessage(statusCode),
+                null,
+                statusCode,
+                null
+            );
+        }
+        try {
+            // ProblemDetails doesn't have a default constructor, and we can't use JSON mapping annotations because
+            // the JSON service is an abstraction over JSON-B and Jackson, so we deserialize the JSON object in a Map
+            // and build the ProblemDetails from the Map values.
```
langsamu, line 105: Would it not be possible to use typed JSON deserialization if this class had a default constructor and conventionally named property methods?

I would prefer that to this Map-based logic.

api/src/test/java/com/inrupt/client/HttpStatusTest.java
```diff
@@ -0,0 +1,63 @@
...
+            HttpStatus.StatusMessages.NOT_FOUND.message
+        );
+    }
+
+    @Test
+    void checkHttpStatusSearchUnknownClientError () {
+        assertEquals(
+                HttpStatus.StatusMessages.getStatusMessage(418),
+                HttpStatus.StatusMessages.BAD_REQUEST.message
+        );
```
langsamu, line 41: This test is a perfect example for why I disagree with the current defaulting logic.

That `418` is most certainly _not_ a `Bad Request`.

api/src/main/java/com/inrupt/client/HttpStatus.java
jholleran: Do we really need to do this status codes mapping at all or could you just pass the data on.
```
        BAD_REQUEST(HttpStatus.BAD_REQUEST, "Bad Request"),
        UNAUTHORIZED(HttpStatus.UNAUTHORIZED, "Unauthorized"),
        FORBIDDEN(HttpStatus.FORBIDDEN, "Forbidden"),
        ...
```

Unless you do all status codes there are going to be problems like this. Will your `HttpStatus` be used as API?

NSeydoux: > I might even argue the it is always the server who is responsible for the content of HTTP status codes and phrases, so our library should just honor the ones generated by the server.

I would have preferred that as well, however not all HTTP libraries expose the HTTP status message, in particular `java.net.http.HttpResponse` does not, so the `com.inrupt.client.ResponseInfo` interface implements the smaller common denominator and doesn't expose the status message returned by the server.

From there, I don't think it is possible to expose what the server returned. The current behavior is aligned with https://www.rfc-editor.org/rfc/rfc9110.html#name-status-codes: 

> HTTP status codes are extensible. A client is not required to understand the meaning of all registered status codes, though such understanding is obviously desirable. However, a client MUST understand the class of any status code, as indicated by the first digit, and treat an unrecognized status code as being equivalent to the x00 status code of that class.

Based on this, I would say it is more correct to default to 400 and 500 for client and server status message respectively rather than returning a unique "Unknown" status message.

api/src/main/java/com/inrupt/client/ProblemDetails.java
NSeydoux: Let me give this a try

NSeydoux: Done in https://github.com/inrupt/solid-client-java/pull/1160/commits/7a59d45e5e05d59ad48246a5c77f02d26183325f via adding a package-private mutable data class used to initialize the immutable one.

api/src/main/java/com/inrupt/client/ProblemDetails.java
NSeydoux: With the typed JSON deserialization, we no longer enforce the problem details status is the same as the HTTP response status, except in case of invalid status in the first place.

api/src/main/java/com/inrupt/client/ProblemDetails.java
```diff
@@ -0,0 +1,143 @@
...
+                null,
+                statusCode,
+                null
+            );
+        }
+        try {
+            final ProblemDetailsData pdData = jsonService.fromJson(
+                    new ByteArrayInputStream(body),
+                    ProblemDetailsData.class
+            );
```
NSeydoux, line 119: I tested this with both jackson and jsonb. The test coverage is added in a different PR.

api/src/main/java/com/inrupt/client/ClientHttpException.java
```diff
@@ -20,35 +20,78 @@
...
+    private final String body;
+    private final transient Headers headers;
 
     /**
      * Create a ClientHttpException.
-     * @param problemDetails the {@link ProblemDetails} instance
      * @param message the exception message
+     * @param uri the error response URI
+     * @param statusCode the error response status code
+     * @param headers the error response headers
```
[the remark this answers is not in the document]
NSeydoux: Fixed with https://github.com/inrupt/solid-client-java/pull/1160/commits/5672101d2d368fd6a4f721f52150c0a7553920f8

api/src/main/java/com/inrupt/client/ProblemDetails.java
```diff
@@ -0,0 +1,146 @@
...
+
+    public int getStatus() {
+        return this.status;
+    };
+
+    public URI getInstance() {
+        return this.instance;
+    };
+
+    public static class ProblemDetailsData {
```
NSeydoux, line 81: Unfortunately `JSONB` demands things be public. I think my previous tests failed to capture this because of lack of correct recompilation

api/src/main/java/com/inrupt/client/ProblemDetails.java
NSeydoux: That turned out to be a lie.

api/src/main/java/com/inrupt/client/Response.java
```diff
@@ -152,6 +154,54 @@ public static BodyHandler<Void> discarding() {
...
 
+        /**
+         * Throws on HTTP error using the provided mapper, or apply the provided body handler.
+         * @param handler the body handler to apply on non-error HTTP responses
+         * @param isSuccess a callback determining error cases
+         * @param exceptionMapper the exception mapper
+         * @return the body handler
+         * @param <T> the type of the body handler
+         */
+        public static <T> Response.BodyHandler<T> throwOnError(
```
NSeydoux, line 165: This body handler can be used in the `solid` module, where we already have a specific mapping from HTTP errors to exceptions

api/src/main/java/com/inrupt/client/Response.java
```diff
@@ -152,6 +154,54 @@ public static BodyHandler<Void> discarding() {
...
+        }
+
+        /**
+         * Throws on HTTP error, or apply the provided body handler.
+         * @param handler the body handler to apply on non-error HTTP responses
+         * @param isSuccess a callback determining error cases
+         * @return the body handler
+         * @param <T> the type of the body handler
+         */
+        public static <T> Response.BodyHandler<T> throwOnError(
```
NSeydoux, line 185: This handler can be used in the `jena` and `rdf4j` modules, where we want to throw a generic HTTP exception.

api/src/main/java/com/inrupt/client/ProblemDetails.java
NSeydoux: I lean towards keeping this over the previous `Map`-based approach: https://github.com/inrupt/solid-client-java/pull/1160/commits/490a1f35b63518a790cdf135466b4b738330dc91, but it does add some surface to the public API.

api/src/main/java/com/inrupt/client/ProblemDetails.java
```diff
@@ -0,0 +1,146 @@
...
+        public URI type;
+        public String title;
+        public String details;
+        public int status;
+        public URI instance;
+    }
+
+    private static JsonService getJsonService() {
```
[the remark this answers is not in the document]
NSeydoux: I thought it would be nicer, but after having implemented some of it, it looks quite the same. One of my assumptions which causes the code to look like what it does here is that the SPI lookup is a relatively expensive operation that you only want to do once, so I want to have a clear separation between "this is null because we haven't initialized it" and "this is null because the lookup failed", so that I don't do a systematic lookup in the latter case assuming we might be in the former.

If I have understood what you are describing correctly, having the `Optional` doesn't change much to the overall structure, in particular because a lookup failure results in an exception, and not in null, so the try/catch needs to be there even when initializing the `Optional` (I think)

api/src/main/java/com/inrupt/client/ProblemDetails.java
```diff
@@ -0,0 +1,146 @@
...
+            // JSON mappers map invalid integers to 0, which is an invalid value in our case anyway.
+            final int status = Optional.of(pdData.status).filter(s -> s != 0).orElse(statusCode);
+            return new ProblemDetails(
+                    type,
+                    pdData.title,
+                    pdData.details,
+                    status,
+                    pdData.instance
+            );
+        } catch (IOException e) {
```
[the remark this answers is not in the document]
NSeydoux: Unfortunately no, more specific exceptions are Jackson or JSON-B specific. Our abstraction layer doesn't define a specific exception beyond `IOException` (it might be a future improvement though, in the spirit of the `ClientHttpException`)

api/src/main/java/com/inrupt/client/Response.java
```diff
@@ -152,6 +154,54 @@ public static BodyHandler<Void> discarding() {
...
+                final Response.BodyHandler<T> handler,
+                final Function<Response.ResponseInfo, Boolean> isSuccess
+        ) {
+            final Function<Response.ResponseInfo, ClientHttpException> defaultMapper = responseInfo ->
+                new ClientHttpException(
+                    "An HTTP error has been returned from "
+                        + responseInfo.uri()
```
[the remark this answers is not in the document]
NSeydoux: Good point, fixed with [3a5f981](https://github.com/inrupt/solid-client-java/pull/1160/commits/3a5f9815e08ff2128206ce62f61f6506aeea2d36)

jholleran (approved):

PR #1161, opened 2024-04-14
JCL-402: Refactor HTTP exception constructors

NSeydoux:
The HTTP exception are now built taking a ProblemDetails in.

### DIFF
--- a/solid/src/main/java/com/inrupt/client/solid/SolidClientException.java
+++ b/solid/src/main/java/com/inrupt/client/solid/SolidClientException.java
@@ -22,7 +22,6 @@ package com.inrupt.client.solid;
 
 import com.inrupt.client.ClientHttpException;
 import com.inrupt.client.Headers;
-import com.inrupt.client.ProblemDetails;
 
 import java.net.URI;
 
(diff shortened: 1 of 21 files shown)


solid/src/main/java/com/inrupt/client/solid/SolidClientException.java
langsamu, line 1: Would it not make more sense to move all the relevant fields and properties to the `ClientHttpException`?
It seems to belong there rather on this specialization.

NSeydoux: Indeed, and actually because https://github.com/inrupt/solid-client-java/pull/1160/commits/2ed0a03ba49d8c5fac59e0f01ae37ee3e545798c embeds the JSON service discovery in the `ProblemDetails` class, https://github.com/inrupt/solid-client-java/pull/1160/commits/b27cd0ad6ac83bd10f37bef351e4796300c8a1db directly builds the problem details from the error response content, which means all the deprecations here are no longer necessary.

NSeydoux: Actually these changes make this PR small enough that it can be bundled together with https://github.com/inrupt/solid-client-java/pull/1160

NSeydoux:
I applied the changes here to #1160

NSeydoux:
This is no longer required based on the refactoring of #1160 .

PR #1162, opened 2024-04-14
JCL-402: RDF4J body handlers http error handling

NSeydoux:
This is based on #1159 , which should be reviewed first. (The only dependency between the two is in the mocks)

This deprecates the current methods in `RDF4JBodyHandlers`, and replaces them with similar method throwing an appropriate exception with error details on HTTP error instead of returning an empty dataset. This makes use of the composable throwing body handler now exposed by the `Response` interface of the `api` module. The new method now have `RDF4J` in their name for this not to be a breaking change: `ofModel` becomes `ofRDF4JModel`, etc.

### DIFF
--- a/rdf4j/src/main/java/com/inrupt/client/rdf4j/RDF4JBodyHandlers.java
+++ b/rdf4j/src/main/java/com/inrupt/client/rdf4j/RDF4JBodyHandlers.java
@@ -41,10 +41,6 @@ import org.eclipse.rdf4j.sail.memory.MemoryStore;
  */
 public final class RDF4JBodyHandlers {
 
-    private static Boolean isSuccess(final Response.ResponseInfo responseInfo) {
-        return responseInfo.statusCode() < 300;
-    }
-
     private static Model responseToModel(final Response.ResponseInfo responseInfo) {
         return responseInfo.headers().firstValue("Content-Type")
                 .map(RDF4JBodyHandlers::toRDF4JFormat).map(format -> {
@@ -76,7 +72,7 @@ public final class RDF4JBodyHandlers {
     public static Response.BodyHandler<Model> ofRDF4JModel() {
         return Response.BodyHandlers.throwOnError(
                 RDF4JBodyHandlers::responseToModel,
-                RDF4JBodyHandlers::isSuccess
+                (r) -> Response.isSuccess(r.statusCode())
         );
     }
 
@@ -114,7 +110,7 @@ public final class RDF4JBodyHandlers {
     public static Response.BodyHandler<Repository> ofRDF4JRepository() {
         return Response.BodyHandlers.throwOnError(
                 RDF4JBodyHandlers::responseToRepository,
-                RDF4JBodyHandlers::isSuccess
+                (r) -> Response.isSuccess(r.statusCode())
         );
     }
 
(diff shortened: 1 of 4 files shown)


langsamu (commented):
Same comment as for the Jena proposal.

rdf4j/src/main/java/com/inrupt/client/rdf4j/RDF4JBodyHandlers.java
```diff
@@ -41,45 +41,83 @@
...
+                    }
+                })
+                .orElseGet(() -> new DynamicModelFactory().createEmptyModel());
+    }
+
     /**
      * Populate a RDF4J {@link Model} with an HTTP response.
      *
      * @return an HTTP body handler
+     * @deprecated
```
langsamu, line 65: Point to new like in Jena.

NSeydoux: Fixed with https://github.com/inrupt/solid-client-java/pull/1162/commits/bf051a11728bddf082be51e988e7cb6cef2d7506

rdf4j/src/main/java/com/inrupt/client/rdf4j/RDF4JBodyHandlers.java
```diff
@@ -41,45 +43,91 @@
  */
 public final class RDF4JBodyHandlers {
 
+    private static void throwOnError(final Response.ResponseInfo responseInfo) {
+        if (!Response.isSuccess(responseInfo.statusCode())) {
+            throw new ClientHttpException(
+                    "Could not map to a Jena entity.",
```
NSeydoux, line 49:
```suggestion
                    "Could not map to an RDF4J entity.",
```

rdf4j/src/test/java/com/inrupt/client/rdf4j/RDF4JBodyHandlersTest.java
```diff
@@ -61,14 +63,14 @@ static void teardown() {
...
-    void testOfModelHandlerAsync() throws IOException,
+    void testOfRDF4JModelHandlerAsync() throws IOException,
             InterruptedException, ExecutionException, TimeoutException {
         final Request request = Request.newBuilder()
                 .uri(URI.create(config.get("rdf_uri") + "/oneTriple"))
                 .GET()
                 .build();
 
-        final Response<Model> res = client.send(request, RDF4JBodyHandlers.ofModel()).toCompletableFuture().join();
+        final Response<Model> res = client.send(request, RDF4JBodyHandlers.ofRDF4JModel()).toCompletableFuture().join();
```
jholleran, line 73: When a method is deprecated should we retain the tests for it and remove them only when the deprecated methods are removed?

NSeydoux: You're right, the deprecated method should be covered until removed.

rdf4j/src/main/java/com/inrupt/client/rdf4j/RDF4JBodyHandlers.java
```diff
@@ -41,45 +43,91 @@
  */
 public final class RDF4JBodyHandlers {
 
+    private static void throwOnError(final Response.ResponseInfo responseInfo) {
+        if (!Response.isSuccess(responseInfo.statusCode())) {
+            throw new ClientHttpException(
+                    "Could not map to an RDF4J entity.",
```
jholleran, line 49: The real issue is that the request failed rather been an issue mapping it to a RDF4J entity. Is this message a bit misleading?

jholleran (approved):

PR #1185, opened 2024-04-19
Error early on test setup failure

NSeydoux:


acoburn:
@NSeydoux I believe this can be closed.

PR #1248, opened 2024-05-06
Support protected resources with RDF body handlers

acoburn:
With the current code, it is not possible to use the `ofJenaModel` (and similar) method on protected resources, because the code will throw an exception at the first 401, interfering with the reactive style of the JCL

### DIFF
--- a/core/src/main/java/com/inrupt/client/core/DefaultClient.java
+++ b/core/src/main/java/com/inrupt/client/core/DefaultClient.java
@@ -21,6 +21,7 @@
 package com.inrupt.client.core;
 
 import com.inrupt.client.Client;
+import com.inrupt.client.ClientHttpException;
 import com.inrupt.client.Headers.WwwAuthenticate;
 import com.inrupt.client.Request;
 import com.inrupt.client.Response;
@@ -84,6 +85,16 @@ public final class DefaultClient implements Client {
             .map(token -> httpClient.send(upgradeRequest(request, token), responseBodyHandler))
             // Otherwise perform the regular HTTP authorization dance
             .orElseGet(() -> httpClient.send(request, responseBodyHandler)
+                .handle((res, err) -> {
+                    if (err instanceof ClientHttpException) {
+                        final ClientHttpException ex = (ClientHttpException) err;
+                        if (ex.getStatusCode() == UNAUTHORIZED) {
+                            return new NoBodyResponse<T>(ex.getUri(), ex.getStatusCode(), ex.getHeaders());
+                        }
+                        throw ex;
+                    }
+                    return res;
+                })
                 .thenCompose(res -> {
                     if (res.statusCode() == UNAUTHORIZED) {
                         final List<Challenge> challenges = WwwAuthenticate
--- a/core/src/main/java/com/inrupt/client/core/DefaultClient.java
+++ b/core/src/main/java/com/inrupt/client/core/DefaultClient.java
@@ -105,7 +105,6 @@ public final class DefaultClient implements Client {
                 }));
     }
 
-
     Request upgradeRequest(final Request request, final Credential token) {
         final Request.Builder builder = Request.newBuilder()
             .uri(request.uri())
--- a/core/src/main/java/com/inrupt/client/core/DefaultClient.java
+++ b/core/src/main/java/com/inrupt/client/core/DefaultClient.java
@@ -85,16 +85,7 @@ public final class DefaultClient implements Client {
             .map(token -> httpClient.send(upgradeRequest(request, token), responseBodyHandler))
             // Otherwise perform the regular HTTP authorization dance
             .orElseGet(() -> httpClient.send(request, responseBodyHandler)
-                .handle((res, err) -> {
-                    if (err instanceof ClientHttpException) {
-                        final ClientHttpException ex = (ClientHttpException) err;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/inrupt/client/core/DefaultClient.java
```diff
@@ -126,6 +128,25 @@ Request upgradeRequest(final Request request, final Credential token) {
         return builder.build();
     }
 
+    @SuppressWarnings("unchecked")
```
langsamu, line 131: Could you reduce suppression from method scope to assignment scope.

core/src/main/java/com/inrupt/client/core/DefaultClient.java
```diff
@@ -126,6 +128,25 @@ Request upgradeRequest(final Request request, final Credential token) {
...
+            if (err instanceof ClientHttpException) {
+                final ClientHttpException ex = (ClientHttpException) err;
+                if (ex.getStatusCode() == UNAUTHORIZED) {
+                    return new NoBodyResponse<T>(ex.getUri(), ex.getStatusCode(), ex.getHeaders());
+                }
+            } else if (err.getCause() instanceof ClientHttpException) {
+                final ClientHttpException ex = (ClientHttpException) err.getCause();
+                if (ex.getStatusCode() == UNAUTHORIZED) {
+                    return new NoBodyResponse<T>(ex.getUri(), ex.getStatusCode(), ex.getHeaders());
+                }
```
langsamu, line 143: Duplicate of above. Might be worth extracting for readability.

core/src/main/java/com/inrupt/client/core/DefaultClient.java
```diff
@@ -126,6 +128,25 @@ Request upgradeRequest(final Request request, final Credential token) {
         return builder.build();
     }
 
+    @SuppressWarnings("unchecked")
+    static <T, R extends Throwable> Response<T> recover(final Response<T> res, final Throwable err) throws R {
+        if (err != null) {
```
langsamu, line 133: Inverting this conditional would reduce indentation of whole method.

core/src/main/java/com/inrupt/client/core/DefaultClient.java
```diff
@@ -84,6 +85,7 @@ public <T> CompletionStage<Response<T>> send(final Request request,
             .map(token -> httpClient.send(upgradeRequest(request, token), responseBodyHandler))
             // Otherwise perform the regular HTTP authorization dance
             .orElseGet(() -> httpClient.send(request, responseBodyHandler)
+                .handle((res, err) -> recover(res, err))
```
langsamu, line 88: Could `recover` be a method reference or are they not available in 8?

core/src/main/java/com/inrupt/client/core/DefaultClient.java
```diff
@@ -126,6 +128,25 @@ Request upgradeRequest(final Request request, final Credential token) {
...
 
+    @SuppressWarnings("unchecked")
+    static <T, R extends Throwable> Response<T> recover(final Response<T> res, final Throwable err) throws R {
+        if (err != null) {
+            if (err instanceof ClientHttpException) {
+                final ClientHttpException ex = (ClientHttpException) err;
+                if (ex.getStatusCode() == UNAUTHORIZED) {
+                    return new NoBodyResponse<T>(ex.getUri(), ex.getStatusCode(), ex.getHeaders());
+                }
+            } else if (err.getCause() instanceof ClientHttpException) {
```
langsamu, line 139: The error and its cause are inpected, but the hierarchy is not traversed further.
I'm sure this is accurate vis a vis our architecture, but it feels off.
I think an explanation with inline comments would be good.

core/src/main/java/com/inrupt/client/core/DefaultClient.java
```diff
@@ -126,6 +128,25 @@ Request upgradeRequest(final Request request, final Credential token) {
...
+                final ClientHttpException ex = (ClientHttpException) err.getCause();
+                if (ex.getStatusCode() == UNAUTHORIZED) {
+                    return new NoBodyResponse<T>(ex.getUri(), ex.getStatusCode(), ex.getHeaders());
+                }
+            }
+            throw (R) err;
+        }
+        return res;
+    }
```
langsamu, line 148: This new functionality is quite complex and branching.
Would it not make sense to add tests that highlight its behavior specifically?

core/src/main/java/com/inrupt/client/core/DefaultClient.java
langsamu: I also think an inline comment here would help, just like the explanatory notes above.

acoburn:
An alternative approach is implemented in #1253

NSeydoux:
This is not directly related to this PR in particular, but should we update the integration tests so that this would have been caught before getting merged?

acoburn:
> This is not directly related to this PR in particular, but should we update the integration tests so that this would have been caught before getting merged?

The issue is that there were new methods introduced. Those new methods were not part of the integration tests, and it was only when I changed the old methods in those integration tests to the new methods that I discovered the implementation problem.

acoburn:
Closed in favor of https://github.com/inrupt/solid-client-java/pull/1253

PR #1255, opened 2024-05-10
Add Spring example with authentication

acoburn:
This adds another Spring-based example application. This application does not currently have any CSS styling to speak of, but the Spring API portion uses recommended, idiomatic constructs.

### DIFF
--- a/examples/spring-web/src/main/java/com/inrupt/client/examples/spring/web/model/WebId.java
+++ b/examples/spring-web/src/main/java/com/inrupt/client/examples/spring/web/model/WebId.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package com.inrupt.client.examples.spring.web.model;
+
+import java.net.URI;
+
+public record WebId(URI id) {
+}
--- a/examples/spring-web/src/main/resources/application.yml
+++ b/examples/spring-web/src/main/resources/application.yml
@@ -1,0 +1,19 @@
+logging:
+  level:
+    root: info
+spring:
+  security:
+    oauth2:
+      client:
+        registration:
+          myApp:
+            client-id: https://inrupt.github.io/solid-client-java/clients/springboot-testing.jsonld
+            provider: inrupt
+            scope:
+              - openid
+              - webid
+            client-authentication-method: none
+            authorization-grant-type: authorization_code
+        provider:
+          inrupt:
+            issuer-uri: https://login.inrupt.com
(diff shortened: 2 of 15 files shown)


examples/spring-web/src/main/resources/static/assets/script.js
```diff
@@ -0,0 +1,38 @@
...
+  };
+
+  // Specific API methods
+  const getUser = () => get("/api/webid");
+  const getProfile = () => get("/api/profile");
+  const getResource = (uri) => get(`/api/resource?uri=${uri}`);
+
+  // Load the application data
+  getUser()
+    .then(user => {
```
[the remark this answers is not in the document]
acoburn: async/await added in https://github.com/inrupt/solid-client-java/pull/1255/commits/c2148f18b925d21dfd76c10d92f052fc7fef9e9c

examples/spring-web/src/main/resources/static/assets/script.js
```diff
@@ -0,0 +1,45 @@
...
+    return fetch(uri).then(res => {
+      if (res.ok) {
+        return res.headers.get("Content-Type") === "application/json" ? res.json() : res.text();
+      }
+      throw new Error("Error fetching data");
+    });
+  }
+
+  // Specific API methods
+  async function getUser() {
```
langsamu, line 16:
```suggestion
  function getUser() {
```

No need since no `await`s.
Returns a promise.

examples/spring-web/src/main/resources/static/assets/script.js
```diff
@@ -0,0 +1,45 @@
...
+    return get("/api/webid");
+  }
+  async function getProfile() {
```
langsamu, line 19:
```suggestion
  function getProfile() {
```

examples/spring-web/src/main/resources/static/assets/script.js
```diff
@@ -0,0 +1,45 @@
...
+    return get("/api/profile");
+  }
+  async function getResource(uri) {
```
langsamu, line 22:
```suggestion
  function getResource(uri) {
```

examples/spring-web/src/main/resources/static/assets/script.js
```diff
@@ -0,0 +1,45 @@
+(async function() {
+  // Clear any state in the URL
+  history.pushState({}, "", "/");
+
+  // General API request handler
+  async function get(uri) {
+    return fetch(uri).then(res => {
```
langsamu, line 7: This `then` is what I would replace with an `await`, otherwise no point in the added `async`.

examples/spring-web/src/main/resources/static/assets/script.js
```diff
@@ -0,0 +1,45 @@
...
+    document.getElementById("user").innerHTML = `<a href="/logout">logout</a>`;
+    document.getElementById("webid").value = user.id;
+
+    const profile = await getProfile();
+    document.getElementById("storage").value = profile.storages[0];
+    document.getElementById("load").disabled = false;
+    document.getElementById("load").addEventListener('click', evt => {
+      document.getElementById("fetch-output").value = "Loading...";
+      getResource(new URL(profile.storages[0]))
+          .then(resource => document.getElementById("fetch-output").value = resource)
```
langsamu, line 38: Also `await getResource` instead of this `then`.

langsamu (commented):
I don't want to waste too much time on this simple demo JS, but I think the current mixing of promise methods and asynchronous keywords is not an improvement.

examples/spring-web/src/main/java/com/inrupt/client/examples/spring/web/model/WebId.java
```diff
@@ -0,0 +1,26 @@
...
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+package com.inrupt.client.examples.spring.web.model;
+
+import java.net.URI;
+
+public record WebId(URI id) {
```
NSeydoux, line 25: Should we mention in a README the expected Java version for the examples, or is it unnecessary as we assume examples will run in a relatively modern Java environment?

examples/spring-web/src/main/resources/application.yml
```diff
@@ -0,0 +1,19 @@
...
+      client:
+        registration:
+          myApp:
+            client-id: https://inrupt.github.io/solid-client-java/clients/springboot-testing.jsonld
+            provider: inrupt
+            scope:
+              - openid
+              - webid
+            client-authentication-method: none
+            authorization-grant-type: authorization_code
```
NSeydoux, line 16: Should we add a comment that this value should match the grant types in the Client ID document?

NSeydoux (approved):
This overall looks good to me, although I don't have enough Spring experience to have a strong opinion on the idiomacy of the proposed app.

examples/spring-web/src/main/java/com/inrupt/client/examples/spring/web/model/WebId.java
acoburn: Spring Boot 3 requires Java 17, so we're good.

examples/spring-web/src/main/resources/application.yml
acoburn: I think authorization_code is a baseline assumption, so I don't see a need to add a comment in particular

PR #1273, opened 2024-05-20
Spring: Remove hard-coded values, move to app configuration

acoburn:
This removes some hard-coded values in the sample Spring application, moving those values to the application configuration


edwardsph (approved):
LGTM

PR #1393, opened 2024-07-02
JCL-476: Adjust ProblemDetails API

acoburn:
There are three significant changes to the ProblemsDetails API in this PR:

* `ProblemDetails::getDetails` becomes `ProblemDetails::getDetail`. This aligns the implementation with the RFC
* The `ProblemDetails.DEFAULT_TYPE` value is now a `URI` (instead of a `String`)
* `SolidClientException::fromErrorResponse` has been refactored.
  * The autocloseable `InputStream` is now handled in a try-with-resources block to prevent resource leakage
  * The method accepts a response `URI` so that any relative `URI`s in the `ProblemDetails` body can be resolved correctly
  * The method is made package-private, since it is not used outside of the package and I don't see a reason to make it part of the public API
  * The logic has been reworked to remove duplication and simplify the flow considerably

In addition, the compiler arguments are changed from the old-style `maven.compiler.source` / `maven.compiler.target` to the new approach `maven.compiler.release`. This eliminates a troublesome compiler warning.

### DIFF
--- a/api/src/main/java/com/inrupt/client/ProblemDetails.java
+++ b/api/src/main/java/com/inrupt/client/ProblemDetails.java
@@ -52,10 +52,10 @@ public interface ProblemDetails {
     String getTitle();
 
     /**
-     * The problem details.
-     * @return the details
+     * The problem detail.
+     * @return the detail
      */
-    String getDetails();
+    String getDetail();
 
     /**
      * The problem status code.
--- a/api/src/main/java/com/inrupt/client/ProblemDetails.java
+++ b/api/src/main/java/com/inrupt/client/ProblemDetails.java
@@ -37,7 +37,7 @@ public interface ProblemDetails {
     /**
      * The <a href="https://www.rfc-editor.org/rfc/rfc9457">RFC9457</a> default problem type.
      */
-    String DEFAULT_TYPE = "about:blank";
+    URI DEFAULT_TYPE = URI.create("about:blank");
 
     /**
      * The problem type.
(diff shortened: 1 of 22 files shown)


solid/src/main/java/com/inrupt/client/solid/SolidProblemDetails.java
```diff
@@ -133,51 +138,29 @@ private static JsonService getJsonService() {
...
-                null,
-                statusCode,
-                null
-            );
+        if (jsonService != null
+                && (headers != null && headers.allValues("Content-Type").contains(ProblemDetails.MIME_TYPE))) {
+            try (final InputStream input = new ByteArrayInputStream(body)) {
+                final Data pdData = jsonService.fromJson(input, Data.class);
+                final URI type = Optional.ofNullable(pdData.type)
+                    .map(uri::resolve)
```
NSeydoux, line 155: The same is required for the instance URI

api/src/main/java/com/inrupt/client/ProblemDetails.java
```diff
@@ -52,10 +52,10 @@ public interface ProblemDetails {
     String getTitle();
 
     /**
-     * The problem details.
-     * @return the details
+     * The problem detail.
+     * @return the detail
      */
-    String getDetails();
+    String getDetail();
```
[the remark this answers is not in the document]
acoburn: None of those changes have been released, so no

solid/src/main/java/com/inrupt/client/solid/SolidProblemDetails.java
acoburn: Addressed in https://github.com/inrupt/solid-client-java/pull/1393/commits/89c65a35adc537e415874b4608963482c2966a92

NSeydoux (approved):

PR #1395, opened 2024-07-03
JCL-476: problem details integration tests

acoburn:
This adds the ProblemDetails checks to the integration testing framework.

The challenge is that, when testing against a live ESS server, there are some versions that already support ProblemDetails responses and others that don't. So this code needs to be able to check based on the capability of the server.


acoburn:
[Tests are succeeding](https://github.com/inrupt/solid-client-java/actions/runs/9783828159/job/27013326222#step:6:657) in the ESS dev environment where ProblemDetails responses are supported:

```
[INFO] Inrupt Java Client Libraries - Integration OpenID Tests SUCCESS [02:32 min]
[INFO] Inrupt Java Client Libraries - Integration UMA Tests SUCCESS [06:28 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
```

NSeydoux:
The current feature detection approach works, but I'd prefer a more explicit feature flagging, because if a given environment disables error descriptions for instance, these tests would keep running, while the test coverage would have decreased. I'd suggest using the same approach as request metadata, with a feature flag passed as an environment variable.

edwardsph (approved):
LGTM

PR #1408, opened 2024-07-11
Revert feature detection configuration

acoburn:
A configuration value was added to aid in feature detection for server-side `ProblemDetails` support in https://github.com/inrupt/solid-client-java/pull/1395/commits/103339f65ffd37774debdd9a08f22854baee2af3 (and subsequent commits in #1395) and then adjusted in https://github.com/inrupt/solid-client-java/pull/1396.

However, that has caused CD to fail ever since and prevents CI from succeeding.

This removes the configuration value entirely and relies solely on feature detection.


edwardsph (approved):
LGTM

PR #1473, opened 2024-08-09
Update jackson dependencies in Jena module

acoburn:
The default jackson dependency brought in by Jena 4.x is somewhat old. This ensures that the jackson version aligns with the rest of the codebase


edwardsph (approved):
LGTM

PR #1545, opened 2024-08-26
Consolidate dependabot management

acoburn:
This adjusts the dependabot configuration to group all plugin updates. This should reduce the noise from dependabot updates across the various supported branches.


edwardsph (approved):
LGTM

PR #1559, opened 2024-09-12
Update maven site model

acoburn:
This updates the maven site model to the new structure


acoburn:
Seems that the `<project>` structure is good as-is

PR #1684, opened 2024-11-08
JCL-474: Introduce filter API

acoburn:
This introduces a new `::query(CredentialFilter)` API in the access grant module.


access-grant/src/main/java/com/inrupt/client/accessgrant/Utils.java
```diff
@@ -119,6 +133,41 @@ public static List<Map<String, Object>> getCredentialsFromPresentation(final Map
         return credentials;
     }
 
+    public static int convertPage(final String page) {
+        try {
+            final int p = Integer.parseInt(page);
```
jholleran, line 138: I thought that page was going to be an opaque values and not something a client should parse. Does the client need to do this or should it follow the links only.

acoburn: This is part of the fallback mechanism for cases where a client is using the client paging API over a non-page-able server API (i.e. /derive). Here, in order to satisfy the expectations of the new response object, the client will do some internal "paging". This could certainly be an opaque value, though it's also not "true" paging.

It's worth noting that the response value here is not a URI, but rather a `CredentialFilter<T>` that can be passed to another invocation of `::query`

acoburn: One option would be to create a random internal value as part of a client instance and use that as part of the basis for paging. But that also seemed like over-complicating a fall-back mechanism.

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
```diff
@@ -347,6 +351,135 @@ public CompletionStage<AccessCredentialVerification> verify(final AccessCredenti
...
+            return new CredentialResult<>(items, null, null, null, null);
+        }
+    }
+
+    <T extends AccessCredential> Map<String, CredentialFilter<T>> processFilterResponseHeaders(final Headers headers,
+            final CredentialFilter<T> filter) {
+        final Map<String, CredentialFilter<T>> links = new HashMap<>();
+        final List<String> linkHeaders = headers.allValues("Link");
+        if (!linkHeaders.isEmpty()) {
+            Headers.Link.parse(linkHeaders.toArray(linkHeaders.toArray(new String[0])))
```
acoburn, line 431: Why are there two `linkHeaders.toArray`

access-grant/src/main/java/com/inrupt/client/accessgrant/AccessGrantClient.java
```diff
@@ -347,6 +351,135 @@ public CompletionStage<AccessCredentialVerification> verify(final AccessCredenti
...
+                .forEach(link -> {
+                    final String rel = link.getParameter("rel");
+                    final URI uri = link.getUri();
+                    if (rel != null && uri != null) {
```
acoburn, line 435: Check values of rel. We only need `first`, `last`, `prev` and `next`

access-grant/src/test/java/com/inrupt/client/accessgrant/MockAccessGrantServer.java
```diff
@@ -173,6 +179,65 @@ private void setupMocks() {
...
+                        .withStatus(401)
+                        .withHeader("WWW-Authenticate", "Bearer,DPoP algs=\"ES256\"")));
+
+        wireMockServer.stubFor(get(urlEqualTo("/query?type=SolidAccessRequest&pageSize=5&page=1"))
+                    .atPriority(1)
+                    .withHeader("Authorization", containing("Bearer eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9."))
+                    .willReturn(aResponse()
+                        .withStatus(200)
+                        .withHeader(CONTENT_TYPE, APPLICATION_JSON)
+                        .withHeader("Link", "</query?type=SolidAccessRequest&page=1&pageSize=20>; rel=\"first\"")
```
acoburn, line 230: Might be useful to add some unexpected link headers

access-grant/src/main/java/com/inrupt/client/accessgrant/Utils.java
```diff
@@ -119,6 +133,41 @@ public static List<Map<String, Object>> getCredentialsFromPresentation(final Map
...
+            return Math.max(1, p);
+        } catch (final IllegalArgumentException ex) {
+            // ignore invalid values
+            LOGGER.debug("Invalid page value, defaulting to first page: {}", ex.getMessage());
```
jholleran, line 142: Could this cause an client to get into a continues loop if there was an error and 1 is returned?

acoburn: I'm not sure how this would result in a loop, unless a developer does some extremely odd things with the code they write (i.e. build a while loop, manipulate the next page filter with an random value and then pass that filter into the next `::query` call. But the developer would need to specifically manipulate the data within that loop. An alternative would be to throw an exception.

jholleran: Throwing an exception seems safer here.

access-grant/src/main/java/com/inrupt/client/accessgrant/CredentialResult.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+public class CredentialResult<T extends AccessCredential> {
+
+    private final List<T> items;
+    private final CredentialFilter<T> first;
+    private final CredentialFilter<T> last;
+    private final CredentialFilter<T> prev;
+    private final CredentialFilter<T> next;
+
+    /* package private */
```
acoburn, line 39: Make this public

access-grant/src/main/java/com/inrupt/client/accessgrant/Utils.java
acoburn: This code has been removed now that the fallback mechanism has been removed.

access-grant/src/main/java/com/inrupt/client/accessgrant/Utils.java
acoburn: This code has been removed via https://github.com/inrupt/solid-client-java/pull/1684/commits/9a97e52b0910518bb41d3a412e2516c25a0d507a

PR #1760, opened 2024-12-13
Release 1.3

acoburn:
This includes the commits for the 1.3 release and prepares for the 1.4 development cycle.

Note that the `gradle` module has been removed. This module was causing an error in the release process, and since it is not actually used (and not deployed), removing it is not an issue. (There was an idea that we would add some gradle-related code at one point, but that never materialized)


edwardsph (approved):
LGTM

PR #1810, opened 2025-01-09
Update NVD cache generation

acoburn:
This updates the gradle build of the open vulnerability project, which is currently failing


edwardsph (approved):
LGTM

PR #1890, opened 2025-02-10
Update java 11 configuration to include smallrye config

acoburn:
This removes the special profiles from the various submodules and puts the smallrye-config version config in the parent pom profile so that dependabot updates are more manageable for Smallrye Config


edwardsph (approved):
LGTM

PR #2360, opened 2025-10-03
JCL-480: Renamed field used for AG link to AR

jholleran:


jholleran:
This needs to be changed to allow the client to optionally switch to the new verified request behaviour to keep it backwards compatible.

